PR #2, opened 2020-05-30
Surface Pro 6 also requires the touchscreen inverting

mtdoughty:
The touchscreen on my Surface Pro 6 needed the same quirk adding to invert the touch screen as existing devices.  With this applied multi touch seems to work reasonably well


StollD:
Thank you!

PR #3, opened 2020-06-02
Surface book 2 15 inch does not need Y axis inverted

libanp:
Exciting to see this working! My Surface book 2 15" does not need to have y-axis inverted


StollD:
Thanks!

PR #5, opened 2020-06-07
Improve accuracy, add simple palm detection

tmarkov:
1) I normalize the heatmap so that untouched squares have 0 by subtracting the average. This makes more sense for the weighted average.

2) Rather than finds contacts, and then estimate coordinates separately, I do it at once: Using BFS I find all clusters of adjacent above-threshold squares and consider them a contact. I track both mean and covariance matrix of each cluster. Tracking the whole cluster improves accuracy, and the covariance matrix can be used for palm detection. To do this, the Contact structure now tracks weighted sums and sums of squares of coordinates, rather than simple coordinates.

Notes:
* Currently, I simply ignore palms. Do they need to generate events too?
* When removing the palm, the cluster sometimes becomes small enough to be mistaken for touch. I guess this needs some sort of contact tracking to fix.
* Not sure how to test if fingers close together get merged in the same cluster.


StollD:
> * Currently, I simply ignore palms. Do they need to generate events too?
> 
> * When removing the palm, the cluster sometimes becomes small enough to be mistaken for touch. I guess this needs some sort of contact tracking to fix.

Linux does have a `MT_TOOL_PALM` type for multitouch input events. Currently our input device does not support the `MT_TOOL` event, but we could change that easily. That could also help with tracking the palms:

Instead of ignoring the contact points for the palms, you could add an `IsPalm` attribute to both, the `Contact` struct and the `TouchInput` struct. Because contact points already get a "unique" index assigned to them through finger tracking, you could extend the finger tracker to also carry over that `IsPalm` flag from the last cycle.

Finally, before emitting events simply check that flag, and if it is true, emit `MT_TOOL_PALM`, otherwise emit `MT_TOOL_FINGER`, and let userspace deal with the rest.

I can whip up a patch for adding `MT_TOOL` support to the device later if you want.

processing/heatmap/data.go
```diff
@@ -27,3 +46,36 @@ func (hm Heatmap) Compare(x1 int, y1 int, x2 int, y2 int) bool {
...
+
+func (c *Contact) Cov() (float32, float32, float32) {
+	return (float32(c.XX) - float32(c.X*c.X)/float32(c.W)) / float32(c.W),
+		(float32(c.YY) - float32(c.Y*c.Y)/float32(c.W)) / float32(c.W),
+		(float32(c.XY) - float32(c.X*c.Y)/float32(c.W)) / float32(c.W)
+}
+
+func (c *Contact) Palm() bool {
+	xvar, yvar, _ := c.Cov()
+	return xvar > 2.5 || yvar > 2.5
```
StollD, line 80: I had to change these two to 1.5 to avoid getting spontanous touches during writing with the stylus, because I slightly lifted part of my palm. Normal touches still seemed to work fine.

processing/heatmap/contact.go
```diff
@@ -1,68 +1,60 @@
...
+		}
+		vis[x][y] = true
+		res.Add(x, y, int(v))
+		queue = append(queue, Pair{x + 1, y}, Pair{x - 1, y}, Pair{x, y + 1}, Pair{x, y - 1})
+	}
+	return res
+}
+
+func (hm Heatmap) Contacts(clusters []Contact) int {
+	threshold := byte(10)
```
StollD, line 26: I had to change the threshold to 30 to avoid getting inputs from pressing the stylus onto the display. Normal touches still seemed to work fine.

processing/heatmap/contact.go
```diff
@@ -1,68 +1,60 @@
...
 	c := 0
 
-	if len(contacts) == 0 {
+	if len(clusters) == 0 {
 		return 0
 	}
 
+	vis := make([][]bool, hm.Width)
```
StollD, line 33: We will need to cache this somehow, or store it differently. make in golang will always allocate on the heap, so any recurring heap allocation is going to cost us a lot of performance, especially since we cannot free the memory manually either, we will just trigger the garbage collector a lot more often.

I originally wanted to store all of these caches in the touch processor struct, since that is already allocated on the heap anyways. The heatmap struct is kept on the stack to make the API a bit simpler and nicer, so we couldn't just add the cache there in this case.

I think we should probably just remove the `processing/heatmap` submodule altogether and just have everything under `processing/` again.

The same applies for the `append` in `GetCluster` btw, but solving that will probably be an even bigger mess, because we don't know a definitive size for these arrays.

processing/heatmap/contact.go
tmarkov: Would a threshold of 30 solve that? I'm not sure the context here, but if I tap the screen with the side of the pen I get higher values than 30.

Wouldn't covariance matrix filter be better here - if variances are 0 (single dot), or the ratio of the covariance matrix eigenvalues is large (long, thin strip), then it's probably not a finger but something else.

processing/heatmap/contact.go
tmarkov: For the queue in GetCluster, an array of length 4*Length*Width should be sufficient. I can put a square in the queue at most 4 times: once when processing each of the four neighbors.

processing/heatmap/contact.go
StollD: I didn't mean the side of the stylus, I mean literally the pen tip. And yeah that was solved by increasing the threshold.

Not sure if we would even want to detect anything other than that?

tmarkov: Ah, I see. On my device the pen tip leaves 0s on the heatmap, hence my confusion.

BTW, I'm not sure how Github works. Do I now need to go and push these changes, have you already applied them? Should I click on 'Resolve conversation'?

StollD: I made these changes locally, for testing. So you would have to make these changes as well and push them to your branch.

processing/heatmap/contact.go
```diff
@@ -1,69 +1,60 @@
 package heatmap
 
-func (hm Heatmap) Contacts(contacts []Contact) int {
-	threshold := hm.Average() + 1
+func (hm Heatmap) GetCluster(x int, y int, threshold byte, vis [][]bool) Contact {
+	res := Contact{}
+	queue := []Pair{{x, y}}
+
+	res.Add(x, y, int(hm.Value(x, y)))
```
StollD, line 7: Wouldn't this add the data for (x,y) two times? Since you start with that in the queue, it will be the first thing thats added before checking other points.

Or is that intended and I just dont understand the algorithm?

StollD:
I tried to fix the issues with lifting palms and get rid of most of the (persistent) heap allocations, and I think I got something that works pretty good. The code is here: https://github.com/linux-surface/iptsd/commits/wip/tmarkov

For the palm detection I just implemented what I described earlier: The palm contacts are not ignored but passed on, tracked by the finger tracker, and sent to userspace using `MT_TOOL_PALM`, which is then ignored by libinput. This gets rid of the ghost touches when lifting the palm.

The heatmap is now heap allocated and cached as whole by the touch processor, instead of caching just the data buffer in the main parser. That makes the code cleaner (yay) and allows storing more stuff directly in the heatmap struct (like the visited stuff). So that is now allocated once and never again.

For the queue in `GetClusters` I actually decided to drop it entirely, and try to implement the function recursively. I think managing a queue with a fixed size array (so just by shifting the index around) is going to be a mess, and I didn't really want to allocate 4 times the size of the heatmap either. Stack operations are cheap, and golang has a pretty large stack (it will fail if your stack exceeds 1MB or so), so doing it recursive seemed like a good idea. The contact thats being checked is passed around as a pointer (it is heap allocated already anyways).

@tmarkov Could you take a look at the changes I made and check if you see any issues with them? They work fine for me, but idk if I missed something during rebasing and rewriting some of the stuff.

tmarkov:
The contact tracing and GetClusters seem to work well.

Palm detection seems fine overall, but there's a bug. Sometimes something goes wrong for me. I get a ButtonPress event without the corresponding ButtonRelease, and then touch starts misbehaving.

My guess, a touch gets detected when placing the palm, only later do we figure out it's a palm, and it's never released.

`xev` output:
```
MotionNotify event, serial 38, synthetic NO, window 0x8400001,
    root 0x1a0, subw 0x0, time 18130022, (236,299), root:(958,905),
    state 0x100, is_hint 0, same_screen YES

MotionNotify event, serial 38, synthetic NO, window 0x8400001,
    root 0x1a0, subw 0x0, time 18130023, (236,299), root:(958,905),
    state 0x100, is_hint 0, same_screen YES

ButtonRelease event, serial 38, synthetic NO, window 0x8400001,
    root 0x1a0, subw 0x0, time 18130023, (236,299), root:(958,905),
    state 0x100, button 1, same_screen YES

MotionNotify event, serial 38, synthetic NO, window 0x8400001,
    root 0x1a0, subw 0x0, time 18130523, (765,190), root:(1487,796),
    state 0x0, is_hint 0, same_screen YES

ButtonPress event, serial 38, synthetic NO, window 0x8400001,
    root 0x1a0, subw 0x0, time 18130523, (765,190), root:(1487,796),
    state 0x0, button 1, same_screen YES

[touch starts misbehaving]
```

StollD:
Could you try this patch? This will discard the palm events again, but after they ran through finger tracking. So when a touch is identified as a palm, the index should get set to -1, which means the touch got lifted.

```diff
diff --git a/touch.go b/touch.go
index 8ad3210..386757c 100644
--- a/touch.go
+++ b/touch.go
@@ -14,14 +14,17 @@ func IptsTouchHandleHeatmap(ipts *IptsContext, heatmap *Heatmap) error {
 	for i := 0; i < len(points); i++ {
 		p := points[i]
 
-		tool := MT_TOOL_FINGER
+		touch.Device.Emit(EV_ABS, ABS_MT_SLOT, int32(i))
+
 		if p.IsPalm {
-			tool = MT_TOOL_PALM
+			touch.Device.Emit(EV_ABS, ABS_MT_TRACKING_ID, -1)
+			touch.Device.Emit(EV_ABS, ABS_MT_POSITION_X, 0)
+			touch.Device.Emit(EV_ABS, ABS_MT_POSITION_Y, 0)
+
+			continue
 		}
 
-		touch.Device.Emit(EV_ABS, ABS_MT_SLOT, int32(i))
 		touch.Device.Emit(EV_ABS, ABS_MT_TRACKING_ID, int32(p.Index))
-		touch.Device.Emit(EV_ABS, ABS_MT_TOOL_TYPE, int32(tool))
 		touch.Device.Emit(EV_ABS, ABS_MT_POSITION_X, int32(p.X))
 		touch.Device.Emit(EV_ABS, ABS_MT_POSITION_Y, int32(p.Y))
 	}
```

tmarkov:
Yes, that patch works pretty well.

tmarkov:
The patch fixes that issue. I notice another thing though: if I have my palm on the screen, then all touches get ignored, and only the pen works. That is perfectly fine with me, but is it the intended behavior?

processing/heatmap/data.go
tmarkov: Source updated.

processing/heatmap/contact.go
tmarkov: Updated threshold

StollD:
Yeah, I forgot that all contacts will need multiple iterations in finger tracking, and that `||` I added modifies the original value. So all inputs end up being palms if the first one that is registered is a palm.

Does this fix the issue?

```diff
diff --git a/processing/finger.go b/processing/finger.go
index 3818ce0..1d0679c 100644
--- a/processing/finger.go
+++ b/processing/finger.go
@@ -93,9 +93,10 @@ func (tp *TouchProcessor) FindDuplicates(count int, itr int) bool {
 		}
 
 		last := tp.last[tp.indices[i][itr]]
+		contact := tp.inputs[i].contact
 
 		tp.inputs[i].Index = last.Index
-		tp.inputs[i].IsPalm = tp.inputs[i].IsPalm || last.IsPalm
+		tp.inputs[i].IsPalm = contact.Palm() || last.IsPalm
 
 		duplicates--
 
@@ -139,9 +140,10 @@ func (tp *TouchProcessor) TrackFingers(count int) {
 	 */
 	for i := 0; i < count; i++ {
 		last := tp.last[tp.indices[i][0]]
+		contact := tp.inputs[i].contact
 
 		tp.inputs[i].Index = last.Index
-		tp.inputs[i].IsPalm = tp.inputs[i].IsPalm || last.IsPalm
+		tp.inputs[i].IsPalm = contact.Palm() || last.IsPalm
 	}
 
 	/*
diff --git a/processing/touch.go b/processing/touch.go
index 577a02c..e70d8a1 100644
--- a/processing/touch.go
+++ b/processing/touch.go
@@ -28,6 +28,8 @@ type TouchInput struct {
 	Y      int
 	Index  int
 	IsPalm bool
+
+	contact *Contact
 }
 
 func (ti TouchInput) Dist(other TouchInput) float64 {
@@ -106,18 +108,20 @@ func (tp *TouchProcessor) Inputs(hm *Heatmap) []TouchInput {
 		}
 
 		tp.inputs[i] = TouchInput{
-			X:      int(x * 9600),
-			Y:      int(y * 7200),
-			Index:  i,
-			IsPalm: tp.contacts[i].Palm(),
+			X:       int(x * 9600),
+			Y:       int(y * 7200),
+			Index:   i,
+			IsPalm:  tp.contacts[i].Palm(),
+			contact: &tp.contacts[i],
 		}
 	}
 
 	for i := count; i < tp.MaxTouchPoints; i++ {
 		tp.inputs[i] = TouchInput{
-			X:     0,
-			Y:     0,
-			Index: -1,
+			X:       0,
+			Y:       0,
+			Index:   -1,
+			contact: &tp.contacts[i],
 		}
 	}
```

tmarkov:
Yes, that fixes the problem I mentioned, but it does bring up another one: When writing with the stylus, sometimes touches get detected near the palm and then I get lines like these:
Screenshot lines

I have been experimenting with a patch that would mark all touches near a palm as palms, but it's still not perfect:

```
diff --git a/processing/contact.go b/processing/contact.go
index af5d96b..ae463ec 100644
--- a/processing/contact.go
+++ b/processing/contact.go
@@ -1,5 +1,9 @@
 package processing
 
+import (
+	"math"
+)
+
 const (
 	TOUCH_THRESHOLD = byte(30)
 )
@@ -11,6 +15,7 @@ type Contact struct {
 	YY int
 	XY int
 	W  int
+	is_palm bool
 }
 
 func (hm *Heatmap) GetCluster(x int, y int, res *Contact) {
@@ -57,6 +62,8 @@ func (hm *Heatmap) Contacts(contacts []Contact) int {
 			contacts[c] = Contact{}
 
 			hm.GetCluster(x, y, &contacts[c])
+			//fmt.Println(contacts[c].Mean())
+			//fmt.Println(contacts[c].Cov())
 			c += 1
 
 			if c == len(contacts) {
@@ -77,13 +84,17 @@ func (c *Contact) Add(x int, y int, w int) {
 	c.W += w
 }
 
-func (c *Contact) Join(d Contact) {
-	c.X += d.X
-	c.Y += d.Y
-	c.XX += d.XX
-	c.YY += d.YY
-	c.XY += d.XY
-	c.W += d.W
+func (c *Contact) Near(other Contact) bool {
+	x1, y1 := c.Mean()
+	x2, y2 := other.Mean()
+	dx, dy := math.Abs(float64(x1-x2)), math.Abs(float64(y1-y2))
+
+	vx, vy, _ := other.Cov()
+
+	sx := 3.2 * math.Sqrt(float64(vx)) + 10
+	sy := 3.2 * math.Sqrt(float64(vy)) + 10
+
+	return dx <= sx && dy <= sy
 }
 
 func (c *Contact) Mean() (float32, float32) {
@@ -109,7 +120,18 @@ func (c *Contact) Cov() (float32, float32, float32) {
 	return r1, r2, r3
 }
 
-func (c *Contact) Palm() bool {
-	x, y, _ := c.Cov()
-	return x > 1.5 || y > 1.5
-}
+func GetPalms(contacts []Contact, count int) {
+	for i := 0; i < count; i++ {
+		vx, vy, _ := contacts[i].Cov()
+		if vx >= 1.5 || vy >= 1.5 {
+			contacts[i].is_palm = true
+			for j := 0; j < count; j++ {
+				if j != i && !contacts[j].is_palm {
+					if contacts[j].Near(contacts[i]) {
+						contacts[j].is_palm = true
+					}
+				}
+			}
+		}
+	}
+}
\ No newline at end of file
diff --git a/processing/finger.go b/processing/finger.go
index 1d0679c..0ee429c 100644
--- a/processing/finger.go
+++ b/processing/finger.go
@@ -96,7 +96,7 @@ func (tp *TouchProcessor) FindDuplicates(count int, itr int) bool {
 		contact := tp.inputs[i].contact
 
 		tp.inputs[i].Index = last.Index
-		tp.inputs[i].IsPalm = contact.Palm() || last.IsPalm
+		tp.inputs[i].IsPalm = contact.is_palm || last.IsPalm
 
 		duplicates--
 
@@ -143,7 +143,7 @@ func (tp *TouchProcessor) TrackFingers(count int) {
 		contact := tp.inputs[i].contact
 
 		tp.inputs[i].Index = last.Index
-		tp.inputs[i].IsPalm = contact.Palm() || last.IsPalm
+		tp.inputs[i].IsPalm = contact.is_palm || last.IsPalm
 	}
 
 	/*
diff --git a/processing/touch.go b/processing/touch.go
index e70d8a1..850e59b 100644
--- a/processing/touch.go
+++ b/processing/touch.go
@@ -92,6 +92,7 @@ func (tp *TouchProcessor) Inputs(hm *Heatmap) []TouchInput {
 	}
 
 	count := hm.Contacts(tp.contacts)
+	GetPalms(tp.contacts, count)
 
 	for i := 0; i < count; i++ {
 		x, y := tp.contacts[i].Mean()
@@ -111,7 +112,7 @@ func (tp *TouchProcessor) Inputs(hm *Heatmap) []TouchInput {
 			X:       int(x * 9600),
 			Y:       int(y * 7200),
 			Index:   i,
-			IsPalm:  tp.contacts[i].Palm(),
+			IsPalm:  tp.contacts[i].is_palm,
 			contact: &tp.contacts[i],
 		}
 	}
```

StollD:
That change makes writing in Xournal++ with the palm on the display way better imo, so we should apply it. Although I would prefer if that `GetPalms` function had less nesting (simply inverting some of the if statements should make that better). And `is_palm` should probabaly be `isPalm`, like the other variable names.

That said, it might be a good idea to add a config option for ignoring all events if a palm is detected on the screen, if someone finds the default behaviour of tracking the palms too buggy.

tmarkov:
OK, pushed the patch.

I was thinking, is there a difference between "primary" and "secondary" touch, as in, how does the OS determine at which finger to keep the cursor?

I notice that if I have two fingers, the left one will have the cursor. But if it's the stylus and a finger, the cursor will jump between them. If we can somehow ensure that the stylus always has priority for the cursor, I think we'll solve the problem for the most part - then palm touches would just interrupt writing or trigger a slight zoom, but won't draw those extra lines over everything.

StollD:
What are you using? X.org or Wayland? Because that sounds like an X.org problem to me.

The X server doesn't know what a "stylus" or a "touchscreen" is. The support for these devices was hacked on top of the old X11 protocol, by making them emulate mouse pointers. That would explain the issue you describe: All touches go through the same device, so X11 only has to expose a single pointer to the window manager. The stylus is a different device though, so once you start using that you get two pointers, which confuse the window manager.

Because Wayland was designed in this century, it knows about these different devices and can properly expose them to client applications. At least under Gnome the touchscreen will not draw a cursor, and the stylus will sometimes render as a crosshair instead of a normal cursor, because the application that draws the cursor can tell what the device actually is.

I don't really know how any of the internals work in that regard, but I can say that with Gnome on Wayland, I don't get random lines drawn if the palm detection fails: The zoom in Xournal++ simply goes haywire, or Gnome starts registering random touch gestures that open the application overview etc.

tmarkov:
I use Mate on X.org - mainly for onboard (last I checked, the gnome screen keyboard was very shitty), but also I regularly use X forwarding (much more convenient than VNC).

That said, what do you think the next steps should be here?

I did some testing with fully blocking palms (by changing the 10s here https://github.com/tmarkov/iptsd/blob/master/processing/contact.go#L94 to 1000), but it doesn't resolve the issue. It seems the remaining problems usually arise when I briefly lift the palm when writing to move the hand. 

Unfortunately, I think that'll be more difficult to resolve - perhaps we need to remember palms for a short while after they've been lifted, so that once it's put back down, it's still tracked, and we know right away it's a palm.

Another approach would be to wait for a while after a new contact to see if it'll turn into a palm, but I'm concerned that will also affect normal touches.

StollD:
> That said, what do you think the next steps should be here?

I think this is ready to be merged. 

I did some experiments with blocking the other inputs when a palm is detected too, and while it doesn't solve every issue, I think it can improve usability in some situations, so I would still like to add an option to enable / disable it. I'm just not sure what the default should be. Currently leaning towards enabling it by default, to get maximum usability while we iron out the remaining quirks in the code.

This is the code I used: I didn't modify the distance check, I just implemented the block directly in the function that sends events. Not sure about the implications, but that way we also keep the list of contacts "intact", since it is cached for finger tracking.

```diff
diff --git a/touch.go b/touch.go
index 386757c..2649c83 100644
--- a/touch.go
+++ b/touch.go
@@ -10,13 +10,18 @@ import (
 func IptsTouchHandleHeatmap(ipts *IptsContext, heatmap *Heatmap) error {
 	touch := ipts.Devices.Touch
 	points := touch.Processor.Inputs(heatmap)
+	blocked := false
+
+	for i := 0; i < len(points); i++ {
+		blocked = blocked || points[i].IsPalm
+	}
 
 	for i := 0; i < len(points); i++ {
 		p := points[i]
 
 		touch.Device.Emit(EV_ABS, ABS_MT_SLOT, int32(i))
 
-		if p.IsPalm {
+		if p.IsPalm || blocked {
 			touch.Device.Emit(EV_ABS, ABS_MT_TRACKING_ID, -1)
 			touch.Device.Emit(EV_ABS, ABS_MT_POSITION_X, 0)
 			touch.Device.Emit(EV_ABS, ABS_MT_POSITION_Y, 0)
```

StollD:
Merged! Thank you for all the work on this, @tmarkov!

I'm going to push some smaller cleanups that I didn't want to annoy you with, and add that config option I mentioned. I will make it default to off though, so we can get more input from testers to further tweak the distance check.

StollD:
Also, @tmarkov, you have an SB1 right? Just so I can add that to the list of tested devices.

tmarkov:
Yes, I have SB1.

PR #7, opened 2020-06-15
Improve palm detection

tmarkov:
1. Use the eigenvalues of the covariance matrix to determine the shape of the touch instead of the variance across the axes. This improves accuracy when the touch is diagonal.
2. To avoid re-computation because of the above, store coordinates and covariance matrix eigenvalues and eigenvectors in the Contact struct, and the weighted sums used for their computation in a separate Cluster struct.
3. Lower the thresholds that designate a cluster as palm if the touch is light.
4. Handle thumb touches (long and thin) separately. This allow further lowering of palm thresholds.
5. Check how fast a cluster grows or shrinks. Ignore touches that grow or shrink too fast as unstable. This helps at the beginning when putting a palm down but the touch area is still small.
6. Lower the heatmap value back to 10 and reject pen tip by 0 variance instead, as it helped with 5. Can somebody test if this works, since my pen doesn't generate any heatmap events.

I'm also considering remembering palms for a short while after they've been lifted, so that if you lift it briefly and put it back down it still gets tracked, but let's get this reviewed first.

~~EDIT: This seems to hinders scrolling. Pretty sure it's a bug, not a fundamental issue with the idea, will update when fixed.~~ Now fixed.

### DIFF
--- a/processing/contact.go
+++ b/processing/contact.go
@@ -20,17 +20,30 @@ func Hypot(x float32, y float32) float32 {
 	return float32(math.Hypot(float64(x), float64(y)))
 }
 
+type Cluster struct {
+	X    int
+	Y    int
+	XX   int
+	YY   int
+	XY   int
+	W    int
+	MaxV int
+}
+
 type Contact struct {
-	X      int
-	Y      int
-	XX     int
-	YY     int
-	XY     int
-	W      int
+	X      float32 // center
+	Y      float32
+	Ev1    float32 // covariance matrix eigenvalues
+	Ev2    float32
+	qx1    float32 // covariance matrix eigenvectors
+	qy1    float32
+	qx2    float32
+	qy2    float32
+	MaxV   int
 	isPalm bool
 }
 
-func (hm *Heatmap) GetCluster(x int, y int, res *Contact) {
+func (hm *Heatmap) GetCluster(x int, y int, res *Cluster) {
 	v := hm.Value(x, y)
 
 	if v < TOUCH_THRESHOLD {
@@ -61,6 +74,8 @@ func (hm *Heatmap) Contacts(contacts []Contact) int {
 		hm.Visited[i] = false
 	}
 
+	var cluster Cluster
+
 	for x := 0; x < hm.Width; x++ {
 		for y := 0; y < hm.Height; y++ {
 			if hm.Value(x, y) < TOUCH_THRESHOLD {
@@ -71,12 +86,12 @@ func (hm *Heatmap) Contacts(contacts []Contact) int {
 				continue
 			}
 
-			contacts[c] = Contact{}
+			cluster = Cluster{}
 
-			hm.GetCluster(x, y, &contacts[c])
-			vx, vy, _ := contacts[c].Cov()
+			hm.GetCluster(x, y, &cluster)
+			contacts[c].GetFromCluster(cluster)
 			// ignore 0 variance contacts
-			if vx > 0 && vy > 0 {
+			if contacts[c].Ev2 > 0 {
 				c += 1
 			}
 
@@ -89,16 +104,19 @@ func (hm *Heatmap) Contacts(contacts []Contact) int {
 	return c
 }
 
-func (c *Contact) Add(x int, y int, w int) {
+func (c *Cluster) Add(x int, y int, w int) {
 	c.X += w * x
 	c.Y += w * y
 	c.XX += w * x * x
 	c.YY += w * y * y
 	c.XY += w * x * y
 	c.W += w
+	if c.MaxV < w {
+		c.MaxV = w
+	}
 }
 
-func (c *Contact) Mean() (float32, float32) {
+func (c *Cluster) Mean() (float32, float32) {
 	fX := float32(c.X)
 	fY := float32(c.Y)
 	fW := float32(c.W)
@@ -106,7 +124,7 @@ func (c *Contact) Mean() (float32, float32) {
 	return fX / fW, fY / fW
 }
 
-func (c *Contact) Cov() (float32, float32, float32) {
+func (c *Cluster) Cov() (float32, float32, float32) {
 	fX := float32(c.X)
 	fY := float32(c.Y)
 	fXX := float32(c.XX)
@@ -121,7 +139,7 @@ func (c *Contact) Cov() (float32, float32, float32) {
 	return r1, r2, r3
 }
 
-func (c *Contact) Eigenvalues() (float32, float32) {
+func (c *Cluster) Eigenvalues() (float32, float32) {
 	vx, vy, cv := c.Cov()
 	sqrtd := Sqrt(((vx-vy)*(vx-vy) + 4*cv*cv))
 
@@ -131,48 +149,55 @@ func (c *Contact) Eigenvalues() (float32, float32) {
 	return r1, r2
 }
 
+func (c *Contact) GetFromCluster(cluster Cluster) {
+	c.X, c.Y = cluster.Mean()
+	vx, vy, cv := cluster.Cov()
+
+	sqrtd := Sqrt(((vx-vy)*(vx-vy) + 4*cv*cv))
+
+	c.Ev1 = (vx + vy + sqrtd) / 2
+	c.Ev2 = (vx + vy - sqrtd) / 2
+
+	c.qx1 = vx + cv - c.Ev2
+	c.qy1 = vy + cv - c.Ev2
+	d1 := Hypot(c.qx1, c.qy1)
+	c.qx1 /= d1
+	c.qy1 /= d1
+
+	c.qx2 = vx + cv - c.Ev1
+	c.qy2 = vy + cv - c.Ev1
+	d2 := Hypot(c.qx2, c.qy2)
+	c.qx2 /= d2
+	c.qy2 /= d2
+
+	c.MaxV = cluster.MaxV
+
+	c.isPalm = false
+}
+
 func (c *Contact) PCA(x float32, y float32) (float32, float32) {
-	vx, vy, cv := c.Cov()
-
-	l1, l2 := c.Eigenvalues()
-
-	qx1 := vx + cv - l2
-	qy1 := vy + cv - l2
-	d1 := Hypot(qx1, qy1)
-	qx1 /= d1
-	qy1 /= d1
-
-	qx2 := vx + cv - l1
-	qy2 := vy + cv - l1
-	d2 := Hypot(qx2, qy2)
-	qx2 /= d2
-	qy2 /= d2
-
-	return qx1*x + qx2*y, qy1*x + qy2*y
+	return c.qx1*x + c.qx2*y, c.qy1*x + c.qy2*y
 }
 
 func (c *Contact) Near(other Contact) bool {
-	x1, y1 := c.Mean()
-	x2, y2 := other.Mean()
-	dx, dy := other.PCA(x1-x2, y1-y2)
+	dx, dy := other.PCA(c.X-other.X, c.Y-other.Y)
 	dx, dy = Abs(dx), Abs(dy)
 
-	vx, vy := other.Eigenvalues()
-	dx /= 3.2*Sqrt(vx) + 5
-	dy /= 3.2*Sqrt(vy) + 5
+	dx /= 3.2*Sqrt(other.Ev1) + 8
+	dy /= 3.2*Sqrt(other.Ev2) + 8
 
 	return dx*dx+dy*dy <= 1
 }
 
 func GetPalms(contacts []Contact, count int) {
 	for i := 0; i < count; i++ {
-		vx, vy := contacts[i].Eigenvalues()
+		vx, vy, maxV := contacts[i].Ev1, contacts[i].Ev2, contacts[i].MaxV
 
-		if vx < 1.0 && vy < 1.0 { // Regular touch
+		if vx < 0.6 || vx < 1.0 && maxV > 80 { // Regular touch
 			continue
 		}
 
-		if vx < 3.0 && vx/vy > 1.8 && vx/vy < 3.4 { // Thumb
+		if (vx < 1.25 || vx < 3.5 && maxV > 90) && vx/vy > 1.8 { // Thumb
 			continue
 		}
 
--- a/processing/contact.go
+++ b/processing/contact.go
@@ -8,6 +8,18 @@ const (
 	TOUCH_THRESHOLD = byte(30)
 )
 
+func Abs(x float32) float32 {
+	return float32(math.Abs(float64(x)))
+}
+
+func Sqrt(x float32) float32 {
+	return float32(math.Sqrt(float64(x)))
+}
+
+func Hypot(x float32, y float32) float32 {
+	return float32(math.Hypot(float64(x), float64(y)))
+}
+
 type Contact struct {
 	X      int
 	Y      int
@@ -82,19 +94,6 @@ func (c *Contact) Add(x int, y int, w int) {
 	c.W += w
 }
 
-func (c *Contact) Near(other Contact) bool {
-	x1, y1 := c.Mean()
-	x2, y2 := other.Mean()
-	dx, dy := math.Abs(float64(x1-x2)), math.Abs(float64(y1-y2))
-
-	vx, vy, _ := other.Cov()
-
-	sx := 3.2*math.Sqrt(float64(vx)) + 10
-	sy := 3.2*math.Sqrt(float64(vy)) + 10
-
-	return dx <= sx && dy <= sy
-}
-
 func (c *Contact) Mean() (float32, float32) {
 	fX := float32(c.X)
 	fY := float32(c.Y)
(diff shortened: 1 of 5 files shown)


tmarkov:
Is it possible for the daemon to cause the module to hang?

In the last hour, after I started using this version, I got two module hangs (lost touch and had to reload the module). I haven't experienced module hangs in weeks. Could the daemon changes have something to do with that, or is it random?

processing/contact.go
```diff
@@ -5,20 +5,45 @@ import (
 )
 
 const (
-	TOUCH_THRESHOLD = byte(30)
+	TOUCH_THRESHOLD = byte(10)
 )
 
+func Abs(x float32) float32 {
```
StollD, line 11: Can you move these utility functions to `utils.go`?

processing/contact.go
```diff
@@ -59,10 +86,14 @@ func (hm *Heatmap) Contacts(contacts []Contact) int {
 				continue
 			}
 
-			contacts[c] = Contact{}
+			cluster = Cluster{}
```
StollD, line 89: Can we merge declaration and assignments for this variable? i.e. instead of having `var cluster Cluster` above, simply do `cluster := Cluster{}` here. Its stack allocated, so it will be thrown away once the loop continues, so we don't need to worry about memory usage.

StollD:
I gave this a quick try, and it is pretty awesome! I was writing some stuff down in Xournal++ earlier today, and I had to enable the "block on palm" option to not get random touches every few seconds. With this, everything worked perfectly, I got maybe 2 random inputs in 2 minutes of writing.

The code looks good to me. I left some minor nitpicks, but there should be no major issue.

Thanks again for your work!

StollD:
> Is it possible for the daemon to cause the module to hang?
> 
> In the last hour, after I started using this version, I got two module hangs (lost touch and had to reload the module). I haven't experienced module hangs in weeks. Could the daemon changes have something to do with that, or is it random?

It shouldn't have anything to do with the changes to the daemon. You didn't change how data is read from IPTS, the only thing that could have changed is how much time passes between reads. And I don't think that is a problem for IPTS. So this sounds more like a random problem.

Can you post your dmesg when the module hangs again? I will try using this a bit more tomorrow too, so if I get any module hangs I will look into them.

tmarkov:
Made a clean-up commit with the requested changes. Also renamed `isPalm` to `IsPalm` in `Contact` since it was `IsPalm` in `TouchInput`.

Here's journal extract from around when I got one of the hangups:
[journal.txt](https://github.com/linux-surface/iptsd/files/4782993/journal.txt)

I see my `rmmod` and `insmod`, but nothing else IPTS-related there unfortunately. Screen simply stopped responding. I restarted iptds, but that didn't fix the issue. So I reloaded the module.

BTW, two more general questions:
How can I make this module preferred over the in-tree one?
What remains to be done before this can become the default over the single-touch module?

tmarkov:
Just got a third ipts hang. Still nothing in journal (`wpa_supplicant[567]: wlp2s0: CTRL-EVENT-SIGNAL-CHANGE above=1 signal=-60 noise=-92 txrate=78000
` and similar messages only for the past five minutes).

Does the uapi module have a debug mode?

touch.go
```diff
@@ -23,6 +23,10 @@ func IptsTouchHandleHeatmap(ipts *IptsContext, heatmap *Heatmap) error {
 
 		touch.Device.Emit(EV_ABS, ABS_MT_SLOT, int32(i))
 
+		if p.Index != -1 && !p.IsStable {
```
tmarkov, line 26: Perhaps this outright skipping the event and emitting nothing can, in some corner case, cause the module to hang? Though it can't join the `if p.IsPalm || blocked {` part, because we don't want button release events if a person suddenly presses harder.

tmarkov:
And four hangs. Looks like this patch is somehow causing it after all. Journal, again nothing:
```
jun 16 00:35:36 tmarkov-surface systemd[1]: Started Network Manager Script Dispatcher Service.
jun 16 00:35:36 tmarkov-surface audit[1]: SERVICE_START pid=1 uid=0 auid=4294967295 ses=4294967295 msg='unit=NetworkManager-dispatcher comm="sy>
jun 16 00:35:36 tmarkov-surface kernel: audit: type=1130 audit(1592256936.809:430): pid=1 uid=0 auid=4294967295 ses=4294967295 msg='unit=Networ>
jun 16 00:35:46 tmarkov-surface systemd[1]: NetworkManager-dispatcher.service: Succeeded.
jun 16 00:35:46 tmarkov-surface audit[1]: SERVICE_STOP pid=1 uid=0 auid=4294967295 ses=4294967295 msg='unit=NetworkManager-dispatcher comm="sys>
jun 16 00:35:46 tmarkov-surface kernel: audit: type=1131 audit(1592256946.782:431): pid=1 uid=0 auid=4294967295 ses=4294967295 msg='unit=Networ>
jun 16 00:44:24 tmarkov-surface mate-screensave[863]: Source ID 888 was not found when attempting to remove it
jun 16 00:45:06 tmarkov-surface rtkit-daemon[684]: Supervising 4 threads of 2 processes of 1 users.
jun 16 00:45:06 tmarkov-surface rtkit-daemon[684]: Supervising 4 threads of 2 processes of 1 users.
jun 16 00:45:30 tmarkov-surface rtkit-daemon[684]: Supervising 4 threads of 2 processes of 1 users.
jun 16 00:45:30 tmarkov-surface rtkit-daemon[684]: Supervising 4 threads of 2 processes of 1 users.
```

touch.go
StollD: No, that can't be the cause. The only interaction with the kernel module happens by reading data from it. No data is sent back. The code you quoted interacts with the uinput module / subsystem of the kernel, not with the ipts module.

StollD:
> Does the uapi module have a debug mode?

No, it doesn't. The `debug` parameter on the "old" driver just enabled dumping all data buffers into dmesg, it did not actually enable any additional logging besides that. All errors that are encountered already get logged into dmesg by default.

This makes your issue a bit unusual since there are no error messages. I could imagine that maybe something prevents the driver from sending feedback, but that should be logged as well.

StollD:
> Made a clean-up commit with the requested changes. Also renamed `isPalm` to `IsPalm` in `Contact` since it was `IsPalm` in `TouchInput`.

Thanks!

> How can I make this module preferred over the in-tree one at boot? Will dkms override the in-tree module by default, or...

I think blacklisting the builtin module with `echo "blacklist ipts" | sudo tee /etc/modprobe.d/ipts.conf`, and then installing the uapi version using dkms should work.

> What remains to be done before this can become the default over the single-touch module?

I would definitly like to see some more testing, at least on a Surface Laptop 1 / 2, and on a gen7 device, to see if these parts work without issues. Also, I still need to figure out packaging for debian, and set up the stuff for pushing releases to the package repository. 

Maybe @qzed also has some additional input on that.

touch.go
tmarkov: But if so nothing in this patch could cause it to hang. What does that leave? A restart-persistent bug in the daemon? Perhaps the frame that "breaks" it doesn't get flushed and then the new instance reads it too, but I don't know if that even makes sense.

StollD: Have you tried to go back to the master branch on this repo? And then check if the module continues to hang?

Even if there was bad data that was sent from the module to the daemon, the daemon could not cause the module to hang, because neither does the module know that the data is bad (because it doesnt parse it), nor can the daemon tell the module about it (because no data is sent back to the module). The only thing that could happen is that the daemon crashes, or that it emits invalid data to uinput and causes the UI to go haywire.

The only thing that is somewhat controlled by the daemon is sending feedback, because the module will wait until a buffer has been read completely by the user, before sending feedback. If that check somehow overflows and fails, feedback would never get sent and the module would lock up because no new data would come in. But I have no idea how you could trigger such a bug.

tmarkov: I plan to try the master, but it's not reliably reproducible, so it'll take time.

But what I meant was, the reason that I believe the module hangs and not the daemon is because the hang persists a daemon restart. Bug if it's possible that the module re-read some data, then perhaps it re-reads the problematic data after restart and hangs again.

Can I somehow manually send feedback to test if lacking feedback's the problem?

qzed:
> > How can I make this module preferred over the in-tree one at boot? Will dkms override the in-tree module by default, or...
> 
> I think blacklisting the builtin module with `echo "blacklist ipts" | sudo tee /etc/modprobe.d/ipts.conf`, and then installing the uapi version using dkms should work.

Not sure if that works actually. DKMS will override modules with the same name automatically (so the old one will be replaced by the new one). Blacklisting ipts might actually then blacklist the new one too.

Just took the changes for a spin. If we want to ship this: Can we make it opt-out, i.e. ship IPTS via iptsd by default but somehow allow switching to single-touch/stylus-only without too much hassle? Or alternatively make the palm-detection (more) configurable before we ship it? Right now, my index finger is misclassified as palm quite regularly (e.g. try moving windows, try drawing with the finger in krita). Works well for light touches but fails as soon as I drag a bit harder (e.g. when moving faster). Could also be that the thresholds need to be adapted per device.

Apart from that some comments on the idea (I have to confess that I haven't looked at the implementation in detail yet, so some of my assumptions might be wrong): Using a covariance matrix to model the shape of the touch and via that (length of major/minor axes, i.e. eigenvalues) detect wether it's a palm or not should generally be a good idea (something I also thought of when working on my [prototype](https://github.com/qzed/touch-detection-prototype/), similar concept there for describing the contact point via covariance matrix). Although I think that may not be enough. If you can get it to work, there's a touch heatmap visualizer in https://github.com/linux-surface/intel-precise-touch/tree/feature/fwd-to-userspc (devtool), that might be of some use to check how the palm actually looks like. IIRC the palm touch points are initially (when the palm starts to touch the screen) fairly similar in size to normal touch points. I think an extension to this method would be to also collect the "mass" under the contact area. Low mass and wide spread should then be considered a palm. The threshold for spread should probably depend on the mass, i.e. so that lighter touches have a more restrictive limit on spread.

Stuff that's not really related to palm detection but should be rather considered as points for future optimization:
- A note on the weighted average to detect the center: This can cause issues on the edge of the screen, i.e. when half of the contact point is cut off. Although I noticed that the sensor seems to go a bit beyond the visible screen area, so that might not be that big of a problem (but then would need some sort of calibration to map this to the screen).
- A note on the clustering: This could lead to problems when doing stuff like "pinch to zoom" and the fingers are close together. Not sure if they'd get detected as two fingers, but even then the simple thresholding will shift the weight (and thus the average) towards the point between the two fingers.

StollD:
> Not sure if that works actually. DKMS will override modules with the same name automatically (so the old one will be replaced by the new one). Blacklisting ipts might actually then blacklist the new one too.

Ah, ok. Good to know, thanks. I always just assumed DKMS would install them side by side.

> Just took the changes for a spin. If we want to ship this: Can we make it opt-out, i.e. ship IPTS via iptsd by default but somehow allow switching to single-touch/stylus-only without too much hassle?

We could keep around the old branch, and if someone wants to revert to the old module, they could install it with `sudo make dkms-install`. Alternatively we could have both modules with different names in the kernel, and only autoload one of them, but that sounds like more hassle for us, and probably the same hassle for users as installing via. DKMS.

tmarkov:
@qzed 

I guess we can think about a calibrator, so far I've obviously only been able to calibrate it to my fingers and device. It shouldn't be too difficult. In the meantime, can you make some finger (not thumb) touches, both light and hard with this patch applied?

```patch
Index: processing/contact.go
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
--- processing/contact.go   (revision 2de748df81ec82df39d3fb009a8ecc92625dab59)
+++ processing/contact.go   (date 1592298817869)
@@ -1,5 +1,7 @@
 package processing

+import "fmt"
+
 const (
    TOUCH_THRESHOLD = byte(10)
 )
@@ -72,6 +74,8 @@
            hm.GetCluster(x, y, &cluster)
            contacts[c].GetFromCluster(cluster)

+           fmt.Println(contacts[c].Ev1, contacts[c].Ev2, contacts[c].MaxV)
+
            // ignore 0 variance contacts
            if contacts[c].Ev2 > 0 {
                c += 1
```

> A note on the weighted average to detect the center: This can cause issues on the edge of the screen

Does that actually happen to you? I've considered adjusting like https://stats.stackexchange.com/questions/48897/maximum-likelihood-estimators-for-a-truncated-distribution, but so far it works well for me without.

> A note on the clustering: This could lead to problems when doing stuff like "pinch to zoom" and the fingers are close together.

That is true. Do you think Gaussian mixtures can be implemented efficiently enough, though? iptsd already seems to be using a significant amount of CPU (4% measured by MATE System Monitor, 13% measured by top, not sure why the difference).

qzed:
Touches (index-finger) while dragging/trying to drag windows. Light touches work, harder touches abort the drag at some point.
- [light.log](https://github.com/linux-surface/iptsd/files/4786680/light.log)
- [hard.log](https://github.com/linux-surface/iptsd/files/4786679/hard.log)

> > A note on the weighted average to detect the center: This can cause issues on the edge of the screen
> 
> Does that actually happen to you? I've considered adjusting like https://stats.stackexchange.com/questions/48897/maximum-likelihood-estimators-for-a-truncated-distribution, but so far it works well for me without.

I haven't been able to test this well. That mostly comes from my earlier experiments with weighted averaging and visualizing the results (something I think we should also do here to be able to qualitatively observe accuracy). However, I noticed that the swipe-in-from-the-left action in Gnome triggers fairly inconsistently, which could be caused by something like this. Again, we'd probably need some tool to visualize that to inspect that better and tell if that's a real issue.

> > A note on the clustering: This could lead to problems when doing stuff like "pinch to zoom" and the fingers are close together.
> 
> That is true. Do you think Gaussian mixtures can be implemented efficiently enough, though? iptsd already seems to be using a significant amount of CPU (4% measured by MATE System Monitor, 13% measured by top, not sure why the difference).

Yeah, that's one thing I worry about with that approach. On the other hand, the majority of the code can be vectorized fairly well. If we combine that with a technique for fast approximate exponentiation (e.g. something like https://stackoverflow.com/questions/47025373/fastest-implementation-of-the-natural-exponential-function-using-sse) and keep the stencil size fairly small (I'd say 5x5 for symmetricity, but 4x4 probably vectorizes a lot better), we should be able to keep the cost down. We might even get away with only two outer (GMM) iterations (the only downside to that should be loss of accuracy when two points are close, and we could also do iterations dynamically based on distance and number of contacts). Another thing is cache/memory access, since stencils can be linearized in memory for better access, it could be possible to eliminate cache misses after the first access. All in all I think it should be possible, but might need some optimizations.

tmarkov:
Thanks for the data. Looks like the heatmap scaling is different across devices - for me these heatmap values represent pretty much the lightest touches I can make, and the palm thresholds depend on how light the touch is.

About the 5x5, I have two concerns about it: I actually tried using a 5x5 to estimate the mean (without looking at std and palm detection), and it was inaccurate. I was mis-touching small elements much more often than I am now. And while I haven't tried estimating standard deviation based on it, I fear it will be much more unstable than mean, being quadratic. As an extreme example, variance explodes to infinity in case the 5x5 is uniform.

qzed:
> About the 5x5, I have two concerns about it: I actually tried using a 5x5 to estimate the mean (without looking at std and palm detection), and it was inaccurate. I was mis-touching small elements much more often than I am now.

The idea on that approach was constructing an SLE to fit a Gaussian. In this case, you don't need to "collect all the weight" as you have to with a normal weighted average. The more you have, the better with respect to noise, but for ideal noise-free data, a couple of points should be enough. That's the reason why it will also work with cut-off data (i.e. at edges and corners). Although I'm quite interested in how this will handle noise. According to some papers, using a weighted least squares penalizer (instead of plain least squares) combined with a couple of iterations should work quite well (although those papers were all describing different use-cases), but there are also extensions to the technique.

> And while I haven't tried estimating standard deviation based on it, I fear it will be much more unstable than mean, being quadratic. As an extreme example, variance explodes to infinity in case the 5x5 is uniform.

The last part is true. The stencil area should always be larger than the object touching the screen (or at least large enough to detect the curvature in the object, after all fingers don't really look like Gaussians if fully pressed on the screen), otherwise this won't work well. So it's not something that will work reasonably well for describing palm contacts. It will identify palm contacts though: If the variance explodes, we know that the area is largely uniform and thus the contact area is large (i.e. larger than the stencil). And if the contact area is large it's very likely not a finger.

If the area is going to be a problem, we can increase the width of the stencil and still keep cost down by selecting pixels based on some pattern (e.g. checkerboard). Again, in an ideal scenario, there shouldn't be any loss due to the SLE, in reality there's probably a loss if we use the wrong stencil pattern for small touch points. We could consider switching stencil pattern based on the initial size estimate (e.g. stencil with more spread for larger contacts). Although that's more of a performance optimization, initially, we should probably just increase the stencil size until all normal touches get recognized properly. After that we can look at reduction of cost.

All in all I think it should work, but this needs some testing.

tmarkov:
I haven't had any hangs in the last three days. @StollD have you had any? If not, I think we can assume it's not caused by this.

StollD:
I haven't used it very intensively, but I didn't have any hangs at all with this.

StollD:
Merged this, since it doesn't seem to cause any issues and works as intented. Thank you again for your work!

PR #16, opened 2020-11-04
WIP: Cone-based palm rejection

tmarkov:
This prototypes a cone-based palm-rejection. When the pen is down or hovering, a 60-degree cone is projected towards a dynamically-determined direction (30 degrees to either side of the direction vector). Touches in the cone are marked as palms.

1) There is only one cone - if there are multiple styluses it will work poorly.
2) Cone direction is determined based on location of palms detected by the shape-detection algorithm. The direction is a weighted average based on previously detected palms. The weight decays with 1 second half-life.
3) Touches in the cone are marked as palms after the the palm proximity detection, so nearby touches will not be marked as palms for proximity.
4) The cone is cut at a specific distance. Touches further than that distance remain touches, but palms further away are used to update direction.


src/touch-processing.c
```diff
@@ -161,5 +172,99 @@ void iptsd_touch_processing_free(struct iptsd_touch_processor *tp)
...
+			if (contacts[j].is_palm)
+				continue;
+
+			if (contact_near(contacts[j], contacts[i]))
+				contacts[j].is_palm = true;
+		}
+	}
+	
+	for (int i = 0; i < count; i++) {
+		if (!contacts[i].is_palm){
```
StollD, line 264: Please add a space between the `)` and the `{`

src/touch-processing.c
```diff
@@ -161,5 +172,99 @@ void iptsd_touch_processing_free(struct iptsd_touch_processor *tp)
...
+			if(iptsd_touch_rejection_cone_is_inside(cone, &contacts[i], timestamp)) {
```
StollD, line 265: Please add a space between the `if` and the `(`. Also, please dont use braces for single statements.

src/utils.c
```diff
@@ -8,6 +8,7 @@
 #include <sys/ioctl.h>
 #include <sys/stat.h>
 #include <unistd.h>
+#include <time.h>
```
StollD, line 11: See above, please sort includes alphabetically.

src/utils.c
```diff
@@ -69,3 +70,9 @@ void iptsd_utils_err(int err, const char *file,
 	va_end(args);
 }
 
+unsigned iptsd_utils_msec_timestamp(void)
+{
+	static struct timespec t;
+	clock_gettime(CLOCK_MONOTONIC, &t);
+	return (unsigned)t.tv_sec * 1000 + t.tv_nsec / 1000000;
+}
```
StollD, line 78: Can you add a newline at the end of the file?

StollD (changes requested):
Thank you! I haven't tried the code yet, but just from reading the diff, here are some (admittedly nitpicky) comments.

src/touch-processing.c
```diff
@@ -161,5 +172,99 @@ void iptsd_touch_processing_free(struct iptsd_touch_processor *tp)
 
 	if (tp->indices)
 		free(tp->indices);
+
+}
+
+void iptsd_touch_rejection_cone_set_tip(
+	struct iptsd_touch_processor *tp, int x, int y, unsigned timestamp)
```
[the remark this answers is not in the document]
tmarkov: How many characters is the line? 80?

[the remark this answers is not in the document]
StollD: Yeah, 80. I basically try to follow the [kernel coding style](https://www.kernel.org/doc/html/v5.8/process/coding-style.html) as much as possible.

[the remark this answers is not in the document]
tmarkov: Timestamps do overflow 32-bit ints, but I'm taking the difference and that should be correct anyway. I guess it wouldn't really hurt to use 64-bit regardless.

src/utils.c
tmarkov: I... think I fixed that? I don't know why it still shows it like that here and doesn't mark it outdated.

tmarkov:
I've fixed up the issues above.

tmarkov:
Actually, it looks like there IS a problem when the cone struct is not dynamically allocated: cone values get zeroed out all the time (not through the update functions). I'm not sure why that happens, but it kind of makes the whole thing useless. Interesting that this does not happen with dynamic allocation.

tmarkov:
It seems that making `struct iptsd_touch_rejection_cone` a part of `iptsd_touch_processor` instead of dynamically allocated, causes it to get zeroed out occasionally (not through the update functions). Not sure exactly why that is, though.

StollD:
So I tried this and it seems to work fine for me. Even with HEAD it works, but I didnt test it for long, so maybe I didn't run into the issue you described yet.

A few questions / ideas though:

* Could we get rid of the timestamps by adding a `bool active;` field to the rejection cone struct, and changing that depending on the value of the proximity bit? So if the stylus is in proximity, the cone gets activated, and when it moves out it gets disabled immideately.
* Could we store one rejection cone per stylus and check all of the active ones once a touch comes in? We can't really test it, but I think it would still be nice if we at least theoretically supported multiple styli.

Also, I think we should move the code for calculating the rejection cone into a seperate, generic file, like contact.c. I.E. removing the iptsd_ prefixes and operating on `struct rejection_cone` (or just `struct cone` maybe?) instead of `struct iptsd_touch_processor`. That way we could keep the function names a bit shorter and dont have to jam all of it into touch-processor.c

tmarkov:
> So I tried this and it seems to work fine for me. Even with HEAD it works, but I didnt test it for long, so maybe I didn't run into the issue you described yet.

I noticed it because it's not rejecting well and by using debug prints. Here's an experiment that should catch it as well:
1) Hold the pen to the screen while touching your palm to set up cone direction.
2) Lift up both.
3) Touch with the pen (no palm thin time).
4) Circle the pen with a finger. For me, it's accepted in all directions.

> Could we get rid of the timestamps by adding a bool active; field to the rejection cone struct, and changing that depending on the value of the proximity bit? So if the stylus is in proximity, the cone gets activated, and when it moves out it gets disabled immediately.

I don't think this will work, since when writing or drawing you lift the pen up sometimes - but you still want palm rejection. So I have it active when the pen is either touching or hovering, so for any pressure.

The second timestamp is used to determine how "fast" to update the cone direction when we detect a new palm (using its shape). I think it's better to have that happen smoothly.

> Could we store one rejection cone per stylus and check all of the active ones once a touch comes in? We can't really test it, but I think it would still be nice if we at least theoretically supported multiple styli.

I think it can work theoretically.
- For rejection, we reject if it's in the cone of any stylus.
- For direction update, we update the direction of the closet stylus.
Do we know which is the currently processed stylus in `iptsd_stylus_handle_data`, to know which cone to move?

> Also, I think we should move the code for calculating the rejection cone into a separate, generic file, like contact.c.

I'd like that, too. I tried it already, but I ended up having .h files wanting to include each other:
- `struct iptsd_touch_processor` needs `struct iptsd_touch_rejection_cone`
- `void iptsd_touch_rejection_cone_set_tip` needs `struct iptsd_touch_processor`
So I either need to put the `void iptsd_touch_rejection_cone_set_tip`, or `struct iptsd_touch_rejection_cone` in `touch_processor.h`, and I figured I don't want to split the touch_processor function across multiple files, so they all went in `touch_processor.h`.

If you have better suggestions on how it can work, I'd appreciate that.

StollD:
> Do we know which is the currently processed stylus in `iptsd_stylus_handle_data`, to know which cone to move?

Yes, the one that we emit events for ;) (`iptsd->devices.active_stylus`). Based on what we figured out about the protocol, it seems that it works a bit like HID: The hardware will send a report containing the serial of the stylus. Until a new serial is sent, all reports come from the stylus with that serial. If iptsd parses a new serial, it will change the active stylus in `iptsd_stylus_change_serial`.

> If you have better suggestions on how it can work, I'd appreciate that.

I would invert the stylus coordinates directly in `iptsd_stylus_handle_data` (since that has direct access to the config). Then `iptsd_touch_rejection_cone_set_tip` doesn't need the touch processor struct, and can work with the rejection cone struct directly.

tmarkov:
> I would invert the stylus coordinates directly in iptsd_stylus_handle_data (since that has direct access to the config). Then iptsd_touch_rejection_cone_set_tip doesn't need the touch processor struct, and can work with the rejection cone struct directly.

I also use `tp->hm.width` and `tp->hm.height` from the touch processor - besides inverting the coordinates, I need to re-scale them. I considered instead re-scaling everything to 7200x9600, but it's not proportional to the screen size, so it'll make the cone differently wide in different directions.

StollD:
Would the width / height that we load from the config work? Thats the actual size of the screen (I am not sure if the heatmap has different sizes on different devices), pulled from the HID descriptors. Currently we only use it to calculate the resolution for the input devices.

https://github.com/linux-surface/iptsd/blob/master/config/surface-book2-13.conf#L9-L10

tmarkov:
OK, I figured out why making the cone static doesn't work. Here, https://github.com/linux-surface/iptsd/blob/e4627e8a3a1e88b452811a1414836e9433ff89e7/src/touch.c#L26, you make a copy of the touch device . Then all the updates are made on the copy and discarded. Do we need that copy for something?

StollD:
Ouch, yeah, we should be getting a pointer to the touch device there. Thanks for spotting it.

tmarkov:
Would you object if I moved all constants to a new constants.h?

StollD:
I'm fine with that. Although my plan was to try and move as many constants as possible into the configuration files, so that users can tweak them if needed.

tmarkov:
I added one rejection cone per stylus and tweaked increased the cone radius.

I tried once more to separate the cone code into its own file, but it just doesn't work out cleanly (I can work around the problems I've mentioned here, but then new problems appear, etc). You can give it a try if you want, but adding multiple cones actually made it even more intertwined with the touch processor, so I think I'll give up on it.

However, there seems to be an outstanding bug which somehow permanently blocks out a part of the screen from touch (but not stylus) until iptsd is restarted. I can't yet reliably reproduce it.

StollD (approved):

StollD:
Thank you for your work (again)! Merged the changes, since they are working as intended.

PR #25, opened 2020-12-05
Add some sanity to build & install scripts

GrayHatter:


StollD:
The service file and the udev rule are both systemd specific, thats why I didn't include them in the meson install. On distros that don't use systemd these will just pollute `/etc` and nothing will ever use them, so it doesn't make sense to forcefully install them with the daemon. Of course the meson file could try to detect the service manager thats used and select the files accordingly but at that point its easier to just install them with a seperate command in the package.

As for the example config, thats a different story. Its obviously the same on all distributions, but its not really required. I added it because people who pull in iptsd through the package repo probably wont read the source code to find out what they can tune.

Also, your changes would put the systemd files under `/etc`, which is meant for local overrides in the systemd world. Config files installed by packages or scripts should go under `/usr` (or `/usr/local` bu I dont know for sure that systemd reads that)

GrayHatter:
Was about to disagree on the sample config file, but on second thought, I agree, removed.

StollD:
This was implemented on master a while ago, so I'm closing this. Sorry that I didnt merge this, but since I wanted to do another change to the meson / systemd stuff, it was easier to do it in one batch. Nonetheless, thank you for the work!

PR #27, opened 2020-12-12
Add tool to capture raw IPTS data

qzed:
Add a tool that can capture raw IPTS data and store it in a binary file for later analysis. This is fairly similar to (and based on) the debug tool that we already have, but instead of dumping it in a human-readable form, this tool will store the unprocessed binary data in a file, which makes it easier to analyze via e.g. python scripts.


qzed:
Replaced by #28

PR #28, opened 2020-12-12
Add debug-tool option to capture raw IPTS data to binary file

qzed:
Add an option to the debug tool that can capture raw IPTS data and store it in a binary file for later analysis. Instead of dumping it in a human-readable form, this option will store the unprocessed binary data in a file, which makes it easier to analyze via e.g. python scripts.

Replaces #27.


StollD (approved):

StollD:
Thank you!

PR #32, opened 2021-01-10
Add openrc init script

ls-r:
What the commits change:
1. Add an init script for the openrc service manager.
2. Change ```meson.build``` so that it can configure and install the new script automatically. The new script is _always_ configured via meson, but whether or not the script is installed depends on ```-Dopenrc=true/false```.
3. Update the systemd portion of ```meson.build``` so that the service file is also always configured. As with the openrc script, only installation is controlled by ```-Dsystemd=true/false```. This change allows the daemon to be more easily packaged for Gentoo.

Notes:
- I considered merging the systemd and openrc options into one option (say ```-Dservice_manager``` for example) which would be an array instead of a boolean. This might help in the future if someone submits a patch for another service manager because you could just add it to the array of options. Let me know if that sounds like a good idea and I'll update the PR.

### DIFF
--- a/etc/openrc/iptsd.initd.in
+++ b/etc/openrc/iptsd.initd.in
@@ -1,0 +1,15 @@
+#!/sbin/openrc-run
+
+description="Intel Precise Touch & Stylus daemon"
+command=@bindir@/iptsd
+
+# Add any arguments to pass to iptsd here
+command_args=""
+
+# Set the pid file location here
+# ${RC_SVCNAME} is set to the service name ("iptsd") when the script is run
+pidfile="/run/${RC_SVCNAME}.pid"
+
+# Set command_background to true if the daemon does not create a pidfile and
+# does not background itself. Start-stop-daemon will take care of these jobs.
+command_background="true"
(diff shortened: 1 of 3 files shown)


StollD:
Nice, thank you! One nitpick though, could you structure the meson file like this:
```
configure_file(<systemd>)
configure_file(<openrc>)

if get_option("systemd")
    <do systemd stuff...>
    install_data(<systemd file>)
end

if get_option("openrc")
    <do openrc stuff...>
    install_data(<openrc file + rename>)
end
```
That wouldnt need the temporary variables you used and unify how the systemd and openrc files are generated, which would make everything a bit more clear imho.



> I considered merging the systemd and openrc options into one option (say `-Dservice_manager` for example) which would be an array instead of a boolean. This might help in the future if someone submits a patch for another service manager because you could just add it to the array of options. Let me know if that sounds like a good idea and I'll update the PR.

That seems like a good idea.

ls-r:
Okay, I restructured to meson.build file to fit your model. The structure you suggested is a lot more elegant! Thanks!

I also merged the options for systemd and openrc. Installation options for both are now handled by ```service_manager```. You can give meson a comma-separated list on the command line like ```-Dservice_manager=systemd,openrc```. The default value is set to install systemd only since that's likely the option most users want.

StollD (approved):

StollD:
Thank you!

Issue #38, opened 2021-03-29
Implement better jitter-reduction


danielzgtg:
This was split from https://github.com/linux-surface/intel-precise-touch/issues/4 as recommended. At least based on the comment there, this isn't implemented in `iptsd` yet.

The outcome is described at https://github.com/linux-surface/intel-precise-touch/issues/4#issuecomment-807622599 :

> You can imagine it like the cursor is a hockey puck with no inertia, to which a rope of a fixed length is tied to your finger. The finger is imagined to be a fixed height above the puck, and the puck only moves when the rope is pulled taught. Shaking your hand a tiny bit will only move the puck once since the rope will become loose when you move your finger back, and scrolling in one direction will always move the puck just like a car pulling a trailer. Compared to the old friction algorithm, this one eliminates both the jitter and at the same time scrolls at a smooth 60 FPS

The algorithm is at https://github.com/danielzgtg/ipts-rust/blob/9d264ad1e629f8e17f96b5e44ac3dea28b7ec765/utils/src/pointers/mod.rs#L116-L137 :

```rust
            if min_dist > 256 {
                fn isqrt(num: i32) -> i32 {
                    let mut result = 3;
                    let mut square = 9;
                    while square < num {
                        result += 1;
                        square += result - 1;
                    }
                    result
                }

                let dx = pos.0 as i32 - old_pos.0 as i32;
                let dy = pos.1 as i32 - old_pos.1 as i32;
                let norm = isqrt((dx * dx) + (dy * dy));
                let dx_clip = (dx * 16) / norm;
                let dy_clip = (dy * 16) / norm;
                pos.0 = (pos.0 as i32 - dx_clip) as u32;
                pos.1 = (pos.1 as i32 - dy_clip) as u32;
            } else {
                pos = old_pos;
            }
```

Basically it doesn't move the cursor until the finger is sqrt(256)=16 pixels away. When it does, it moves the cursor along the displacement vector until it is 16 pixels away again.

This filtering algorithm is fast and numerically stable.

bngo11:
@danielzgtg Hi, I was able to get your driver to work. The zoom worked where it would not snap back to previous position. But I found so other issues with it. During the zoom, the image would be shaky/jittery.

Occasionally it would register a different location. We have a login screen with a keypad, sometime pressing on the number one would have the key below it, in this case a 'clear' key, being pressed instead of the '1' key.

[...]

PR #39, opened 2021-05-29
Implement jitter reduction

mklss:
This implements jitter reduction as in #38.

I've used this a few times, and it seems to work. However, there is some unexpected behaviour.
Sometimes, the finger detection seems to fail with `PositionStabilityThreshold = 0`. In fact, the changes seem to negatively affect performance and leads to a bit of sluggishness. Very brief fast touches are ignored. However, I observe very same behaviour when I compile without the changes. So maybe it's the locally compiled executable which is the problem. Are optimizations required (and how would I enable them).


mklss:
@StollD: Do you have time to review/test (and merge) this? If you won't merge, do you know why running iptsd from terminal (even without the changes) the movement is more jerky/it feels like the update frequency is lower. Also, just replacing /usr/bin/iptsd with the build breaks ("Display size is 0!".), so I suppose something special happens for these builds.

StollD:
Sorry I've been really busy the last weeks so this got pushed further and further behind. :/
From a glance over it the code looks good, so I will just merge it before letting you wait another month.

> If you won't merge, do you know why running iptsd from terminal (even without the changes) the movement is more jerky/it feels like the update frequency is lower.

Did you build with optimizations (`meson build -Doptimization=2? meson doesnt enable them by default, while our packages do.

> Also, just replacing /usr/bin/iptsd with the build breaks ("Display size is 0!".),

Local builds will have their prefix set to `/usr/local`. This means that the directory for the config files evaluates to `/usr/local/share/ipts`. The release packages set the prefix to `/usr`, so the path is `/usr/share/ipts`. If you run it directly from the checked out repo it will work, because the program falls back to loading the configs from `./config`.

Either install your local build with `sudo ninja -C build install`, or build with `meson build --prefix=/usr`

tmarkov:
@StollD @mklss The lack of absolute value that you correct in the "use absolute value" commit isn't a bug. https://github.com/linux-surface/iptsd/issues/38#issuecomment-850820715

StollD:
@mklss Would there be any side effects if we reverted the first commit? (https://github.com/linux-surface/iptsd/pull/39/commits/d1c4df95401989abaefeaffa12784c4d3a5b4d56). Given that you dont use the `dev1` and `dev2` variables later on, it should be fine?

PR #50, opened 2022-01-11
Updates for Surface Book3

martin-lueker:
- Add a new heatmap type IPTS_HID_REPORT_HEATMAP_V2
- Import new screen geometry value (for 15") based on observations


martin-lueker:
Hi @StollD, 
   I added two commits here.  In my experience, the first is very much needed for getting MT to work with my SB3 (iptsd/libqzed won't even recognize the heatmaps without it).  

The second just updates the 15" screen geometry to match the observed geometry of my device.   Was the previous geometry a placeholder/guess, or was this a value that had been observed in other 15" SB3's?   In either case, I'd be happy to dump this second commit.

Looking forward to hearing your input.
(and also thanks for all your hard work developping iptsd up to this point!)

StollD:
The first change looks sensible, thank you.

Please revert the second change though. These values don't specify the resolution of the display, but its physical size (in the weird unit of mm * 10, because thats how the HID descriptor reports it). That value is used to un-distort the heatmap, since the aspect ratio of the heatmap and the aspect ratio of the display don't have to line up exactly.

martin-lueker:
Hi @StollD, 
  Thanks for explaining the units here!  The second commit has been expunged from this PR.

StollD (approved):

StollD:
Thank you!

PR #62, opened 2022-07-18
Add config file for Surface Pro X SQ2

qzed:
I assume the Surface Pro X SQ1 uses the same HID device ID as the SQ2.


StollD:
Thank you!

PR #63, opened 2022-07-18
pkg: Allow building Arch Linux package on AArch64

qzed:
Allow building the Arch Linux package on AArch64, i.e., the Surface Pro X.


StollD:
Thank you!

(Also feel free to commit directly to the master branch)

PR #76, opened 2022-09-02
Add support for metadata feature report

quo:
This allows screen size and orientation to be read from the device, so config files are no longer necessary.

`Parser.dim.width/height` are now also initialized on startup, so that the pen works even if the user has not yet touched the screen.

I removed the config files for the SP7, SP7+, SP8, SB3, SL3, SLS and SPX, since I believe those all support the metadata report. But I'm not completely sure about all of them. And there may be more that can be removed?

I didn't look at the debug tools. They probably also need some changes if the config files are removed.

I've only tested on an SP7+.


qzed:
This fails on the SPX: "Tried to read more data than available!".

quo:
Interesting! Can you post the contents of the report?

E.g. with:
```patch
diff --git a/src/ipts/device.cpp b/src/ipts/device.cpp
index 89d178e..190a94d 100644
--- a/src/ipts/device.cpp
+++ b/src/ipts/device.cpp
@@ -7,6 +7,8 @@
 #include <cstddef>
 #include <iterator>
 #include <vector>
+#include <spdlog/spdlog.h>
+#include <spdlog/fmt/bin_to_hex.h>
 
 namespace iptsd::ipts {
 
@@ -117,6 +119,7 @@ std::optional<Metadata> Device::get_metadata()
        report.at(0) = id;
 
        this->get_feature(report);
+       spdlog::info("{}", spdlog::to_hex(report));
 
        std::optional<Metadata> metadata = std::nullopt;
        Parser parser;
```

qzed:
Working on it. Can confirm that it fails parsing the metadata report. That spdlog solution is quite a bit simpler than the for loop I was testing just now :)

qzed:
metadata report:

```
0000: 77 00 00 00 00 00 00 70 00 00 00 00 02 01 2e 00 00 00 44 00 00 00 fd 6a 00 00 53 47 00 00 01 41
0020: 65 cc 43 00 00 00 00 00 00 00 00 00 00 00 00 b6 e0 ca 43 00 00 00 00 00 00 32 43 00 00 36 43 00
0040: 00 34 43 00 00 80 3f 00 00 32 43 00 00 36 43 00 00 34 43 00 00 80 3f 00 00 b4 42 00 00 2b 43 00
0060: 00 c8 42 00 00 a0 41 00 00 2c 43 00 00 31 43 00 00 2f 43 00 00 00 40 00
```

Invocations of parser functions:
```
parser.skip:
  size: 1
  this->size: 120

parser.read
  dest.size: 7
  this->size: 119

parser.sub
  size: 18446744073709551609
  this->size: 112

[21:06:18.987] [error] Tried to read more data than available!
```

quo:
Is it possible the GET_REPORT implementation in your HID driver is broken?

For me the format is identical, except everything is shifted over by one byte, the first byte is the report ID, and the last nul byte is missing.

qzed:
Huh, yeah the report looks a bit odd now that you mention it. Possible. The 0x77 seems to be the length. So before that should be the report ID?

quo:
Right. This is what I get:
```
0000: 06 77 00 00 00 00 00 00 70 00 00 00 00 02 01 2c 00 00 00 40 00 00 00 7a 65 00 00 a7 43 00 00 01
0020: b3 2c ce c3 00 00 00 00 00 f4 ca 46 00 00 00 00 3c 62 c9 c3 00 4e 87 46 00 00 32 43 00 00 36 43
0040: 00 00 34 43 00 00 80 3f 00 00 32 43 00 00 36 43 00 00 34 43 00 00 80 3f 00 00 b4 42 00 00 2b 43
0060: 00 00 c8 42 00 00 a0 41 00 00 2c 43 00 00 31 43 00 00 2f 43 00 00 00 40
```

qzed:
I'll try to fix that up in IPTSd to confirm and then check the driver. I'd say it's pretty likely that that is broken then.

qzed:
Perfect, with that hacked around it works! Thanks! So I guess I'm off fixing that driver then.

qzed:
Last comment from my side on this story: Confirmed kernel driver was broken (still is in the surface duo kernel tree I grabbed it from... so if anyone wants/knows how to report that to MS, feel free). Fixed it (https://github.com/linux-surface/kernel/commit/5517bd52ee40ad889a69192f235730628aec57ab). Now it's working perfectly.

src/daemon/main.cpp
```diff
@@ -57,6 +68,8 @@ static int main(gsl::span<char *> args)
 	spdlog::info("Connected to device {:04X}:{:04X}", device.vendor(), device.product());
 
 	ipts::Parser parser {};
+	if (meta.has_value())
```
StollD, line 71: I am wondering if we could pass the dimensions directly to the DFT processing instead of taking the detour over the parser.

src/daemon/main.cpp
```diff
@@ -47,7 +48,17 @@ static int main(gsl::span<char *> args)
 	auto const _sigint = common::signal<SIGINT>([&](int) { should_exit = true; });
 
 	ipts::Device device {path};
-	config::Config config {device.vendor(), device.product()};
+
+	auto meta = device.get_metadata();
+	if (meta.has_value()) {
+		auto &t = meta->transform;
+		spdlog::info("Metadata: rows={}, columns={}, width={}, height={}, transform={}, unknown={}, {}",
```
StollD, line 55: Not really against this, but I think this would make more sense in the debug tools.

src/ipts/device.cpp
```diff
@@ -81,4 +107,24 @@ void Device::set_mode(bool multitouch)
...
+	if (!id)
+		return std::nullopt;
+
+	std::vector<u8> report(1 + this->descriptor().size(id));
+	report.at(0) = id;
+
+	this->get_feature(report);
+
+	std::optional<Metadata> metadata = std::nullopt;
+	Parser parser;
```
StollD, line 122: I think the structure should be parsed directly (using `ipts::Reader`) instead of using the parser. Its simple enough and doesnt require changing the parser.

src/ipts/parser.hpp
```diff
@@ -59,6 +59,14 @@ class DftWindow {
 	std::array<struct ipts_pen_dft_window_row, IPTS_DFT_MAX_ROWS> y {};
 };
 
+class Metadata {
```
StollD, line 62: I would prefer if this was a structure in `protocol.hpp` and parsed in one go.

src/ipts/protocol.hpp
```diff
@@ -173,4 +179,20 @@ struct [[gnu::packed]] ipts_pen_dft_window_row {
 	i8 zero;
 };
 
+struct ipts_touch_metadata_size {
```
StollD, line 182: These structures should be marked as `[[gnu:packed]]`. I dont actually think thats required, but its probably good to be safe and consistent.

src/ipts/protocol.hpp
```diff
@@ -173,4 +179,20 @@ struct [[gnu::packed]] ipts_pen_dft_window_row {
...
+	u32 rows;
+	u32 columns;
+	u32 width;
+	u32 height;
+};
+
+struct ipts_touch_metadata_transform {
+	float xx, yx, tx;
```
StollD, line 190: This should be `f32`, for consistency with the rest of the code.

src/daemon/main.cpp
quo: Possibly. Ideally we would pass the entire matrix around and use that everywhere, however that would require larger changes.

src/daemon/main.cpp
quo: I'm interested if anyone has different values for the unknown parts, figured this would be easier than asking people to run debug tools.

src/ipts/device.cpp
quo: I considered that. Two reasons I went with this: 1. The structure clearly follows the same format as the HID data. 2. For debugging you might want to dump metadata and data to the same data stream, so this could make it easier to parse that stream (although you would still need to deal with the lack of timestamp).

I'm not happy about the `has_timestamp` flag of course. But I feel the correct way to fix that is to only give the parser the container part of each report instead of the entire HID report.

If you prefer not to put the metadata parsing in `Parser` though, I'll change it.

src/ipts/parser.hpp
quo: That's what I had initially. However I suspect the unknown byte could be a count for the structures after that, due to the lack of alignment, so I split it up. Also the single packed structure led to issues when I wanted to pass the values to spdlog. And (if we keep the parsing in `Parser`) I feel it makes more sense to have `Metadata` next to `StylusData`, etc.

src/ipts/protocol.hpp
quo: Like you say, it's not necessary here. However the only reason I didn't add it is because it made passing the values to spdlog tricky.

src/ipts/protocol.hpp
quo: Right, will change that.

StollD:
Hey, first of all sorry for letting this slip. Life got in the way...

Since this was already working well and the changes I had in mind were somewhat minor, I decided to make them myself to get this merged soon. The branch is here: https://github.com/linux-surface/iptsd/tree/pr-metadata

I would appreciate if you could try out that branch to make sure I didnt break anything.

quo:
No worries. Thanks for looking at it and cleaning it up a bit, looks good.

You missed an equals sign after `height` here, and also in the debug tools:
https://github.com/linux-surface/iptsd/blob/1b28f531dea2dac2b3611b4fb9dc2c5bbc1d4929/src/daemon/main.cpp#L58

I'll have some time to test it over the weekend.

quo:
On an SP7+ with ithc, everything seems to work, including the debug tools.

However, on an SP7 (with ipts master), iptsd fails with "Failed to get feature: Input/output error", and dmesg has "ipts ...: GET_FEATURES timed out!".

StollD:
Thats interesting. That error also happens on my SB2, but it has always done that. I just dismissed it as a bug of the older IPTS implementation where HID was an afterthought.

Did this work on SP7 before?

You could try raising the timeout here: https://github.com/linux-surface/intel-precise-touch/blob/master/src/hid.c#L165
But I dont think that the response should take longer than 5 seconds.

Maybe my code for waiting on the message is just broken?

StollD:
You could also try this patch. That should catch every possible response to the GET_FEATURES command, no matter their actual type.

```diff
diff --git a/src/hid.c b/src/hid.c
index 1d6c6ff..8aa42d4 100644
--- a/src/hid.c
+++ b/src/hid.c
@@ -213,16 +213,16 @@ int ipts_hid_input_data(struct ipts_context *ipts, int buffer)
        if (header->size == 0)
                return 0;
 
-       if (header->type == IPTS_DATA_TYPE_HID)
-               return hid_input_report(ipts->hid, HID_INPUT_REPORT, header->data, header->size, 1);
-
-       if (header->type == IPTS_DATA_TYPE_GET_FEATURES) {
+       if (header->buffer == IPTS_HID2ME_BUFFER) {
                ipts->get_feature_report = header->data;
                complete_all(&on_feature);
 
                return 0;
        }
 
+       if (header->type == IPTS_DATA_TYPE_HID)
+               return hid_input_report(ipts->hid, HID_INPUT_REPORT, header->data, header->size, 1);
+
        if (header->type != IPTS_DATA_TYPE_FRAME)
                return 0;
 ```

**Comment by quo** [comment]
I got it working. Here's a fix for ipts:

```patch
diff --git a/src/hid.c b/src/hid.c
index 1d6c6ff..dba86ac 100644
--- a/src/hid.c
+++ b/src/hid.c
@@ -150,6 +150,7 @@ static int ipts_hid_raw_request(struct hid_device *hid, unsigned char reportnum,
 		}
 	}
 
+	buf[0] = reportnum;
 	ret = ipts_hid_hid2me_feedback(ipts, type, buf, size);
 	if (ret) {
 		dev_err(ipts->dev, "Failed to send hid2me feedback: %d\n", ret);
@@ -172,7 +173,7 @@ static int ipts_hid_raw_request(struct hid_device *hid, unsigned char reportnum,
 		return -EFAULT;
 
 	memcpy(buf, ipts->get_feature_report, size);
-	return 0;
+	return size;
 }
 
 static int ipts_hid_output_report(struct hid_device *hid, __u8 *data, size_t size)
```

Also, the SP7 constantly sends invalid packets, and after 50 of those iptsd will quit. You can use the following to ignore the invalid packets:

```patch
diff --git a/src/ipts/parser.cpp b/src/ipts/parser.cpp
index 91cb8cb..af0bd42 100644
--- a/src/ipts/parser.cpp
+++ b/src/ipts/parser.cpp
@@ -75,6 +75,13 @@ void Parser::parse_hid(Reader reader)
 			this->parse_heatmap_frame(sub);
 			break;
 		case IPTS_HID_FRAME_TYPE_REPORTS:
+			// On SP7 we receive the following data about once per second:
+			// 16 00 00 00 00 00 00
+			//   0b 00 00 00 00 ff 00
+			//     74 00 04 00 00 00 00 00
+			// This causes a parse error, because the "0b" should be "0f".
+			// So let's just ignore these packets:
+			if (reader.size() == 4) return;
 			this->parse_reports(sub);
 			break;
 		case IPTS_HID_FRAME_TYPE_METADATA:
```

StollD:
Thanks, that seems like a good idea. I also added setting the rest of the buffer to zero, I always just assumed that the kernel would already do that (idk why).

Sadly it still doesnt make GET_FEATURES work on SB2, but hey, I've given up on that a long time ago :P 

I merged the pr-metadata branch, sadly github is now confused and didn't mark this PR as merged, so I will close it. Thank you for your work!

quo:
Great, thanks!

> I also added setting the rest of the buffer to zero

Actually, it turns out you only need to send the report number for GET_FEATURE, so instead of `buf[0] = reportnum` you could do `ipts_hid_hid2me_feedback(ipts, type, &reportnum, 1)`. You'd need a separate path for output reports though, and I don't know if those would need `buf[0] = reportnum` (it probably won't hurt).

> Sadly it still doesnt make GET_FEATURES work on SB2

What does the HID descriptor look like on the SB2?

And BTW, you're still missing the equals sign in the metadata logging.

StollD:
> What does the HID descriptor look like on the SB2?

The device itself doesnt provide one, the windows driver and the old Intel one for linux load it from a binary blob. Here is the one from SB2: https://github.com/linux-surface/surface-ipts-firmware/blob/master/firmware/intel/ipts/MSHW0137/vendor_desc.bin

https://gist.github.com/StollD/de3aebcac9602e151c05754d4e9969bb

It is very similar to what the newer device reports, but its missing all the reports from the section with the feature reports for setting mode and getting metadata. My guess is that these are reports that control the windows driver, rather than the device, and that the actual device supports no feature reports at all.

> And BTW, you're still missing the equals sign in the metadata logging.

I actually wrote that commit on my desktop and then did the merge on my laptop... Thanks for reminding me.

PR #104, opened 2023-01-04
Improve scripts and fix error

m1nicrusher:
Remove $ sign to allow copy & paste & run.
Expanded `semanage` codes since `*` won't work.

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,18 +67,10 @@ On Fedora (or any other SELinux enabled distribution) you also need to fix the S
 
 ```bash
 $ sudo semanage fcontext -a -t systemd_unit_file_t -s system_u /usr/lib/systemd/system/iptsd@.service
-$ sudo semanage fcontext -a -t usr_t -s system_u /usr/local/bin/iptsd
-$ sudo semanage fcontext -a -t usr_t -s system_u /usr/local/bin/iptsd-calibrate
-$ sudo semanage fcontext -a -t usr_t -s system_u /usr/local/bin/iptsd-dump
-$ sudo semanage fcontext -a -t usr_t -s system_u /usr/local/bin/iptsd-find-hidraw
-$ sudo semanage fcontext -a -t usr_t -s system_u /usr/local/bin/iptsd-perf
+$ sudo semanage fcontext -a -t usr_t -s system_u '/usr/local/bin/ipts.*'
 
 $ sudo restorecon -vF /usr/lib/systemd/system/iptsd@.service
-$ sudo restorecon -vF /usr/local/bin/iptsd
-$ sudo restorecon -vF /usr/local/bin/iptsd-calibrate
-$ sudo restorecon -vF /usr/local/bin/iptsd-dump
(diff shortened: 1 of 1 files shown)


StollD:
I kinda like having the dollar sign, because it means that you have to somewhat think about what you are doing. You cant just do Ctrl-A and Ctrl-C.

And why are the wildcards not working? Does semanage not support multiple paths? I vaguely remember that semanage also supports wildcards by itself, does it work if the path is wrapped with `''`?

NP-chaonay:
IMHO

> Remove $ sign to allow copy & paste & run.

if that code is not designed to be copy&paste directly, then this may not required, the $ and # is indicate that if that command required root or not

> Expanded semanage codes since * won't work.

it also depend on shell that using, so maybe we may set the defination if the given command line required some specific shell (bash/sh?) or not.

BTW I'm neutral on this because not know much context

m1nicrusher:
> I kinda like having the dollar sign, because it means that you have to somewhat think about what you are doing. You cant just do Ctrl-A and Ctrl-C.

OK then, a valid point.

> And why are the wildcards not working? Does semanage not support multiple paths? I vaguely remember that semanage also supports wildcards by itself, does it work if the path is wrapped with `''`?

I'm using zsh and wildcards expanded the multiple paths into multiple arguments, which semanage then posted an error about too many arguments.

For example if we have `/etc/a` and `/etc/b`,
doing `semanage --flags /etc/*` will convert it into `semanage --flags /etc/a /etc/b` and it will complain about "/etc/b" as an excess argument that is hence not recognised.

(I don't quite remember the detailed error message but it was not until I manually expanded the paths into separate commands that it started to work.)

m1nicrusher:
I have re-added the $ sign at each line of code if that's preferred.
I've tested the wildcard to a brand new install of Fedora with zsh as shell.

This is what I get:
<pre>sudo semanage fcontext -a -t usr_t -s system_u /usr/local/bin/ipts*
usage: semanage [-h]
                {import,export,login,user,port,ibpkey,ibendport,interface,module,node,fcontext,boolean,permissive,dontaudit}
                ...
semanage: error: unrecognized arguments: /usr/local/bin/iptsd-calibrate /usr/local/bin/iptsd-dump /usr/local/bin/iptsd-find-hidraw /usr/local/bin/iptsd-perf /usr/local/bin/iptsd-plot /usr/local/bin/iptsd-show
</pre>

In this case, it does not allow multiple target arguments in single execute.

StollD:
So, I was fairly sure that semanage supported wildcards (passed in single quotes so they are not evaluated by the shell), but after looking it up it seems that it actually supports regular expressions.

So, what about something like this? `sudo semanage fcontext -a -t usr_t -s system_u /usr/local/bin/ipts.*`

m1nicrusher:
> sudo semanage fcontext -a -t usr_t -s system_u '/usr/local/bin/ipts.*'

Looks like it works fine on my machine. This is indeed much neater.

StollD:
Thank you!

PR #106, opened 2023-02-01
Fix some systemd warnings about unescaped arguments

qzed:
Systemd complains about unescaped arguments to the service. So escape them properly.


StollD (approved):

StollD:
Thank you!

PR #110, opened 2023-02-22
Fix Parser::stylus being shadowed and not written to

danielzgtg:
`Parser::parse_stylus_v{1,2}` redeclares `StylusData stylus`, shadowing `Parser::stylus`. The member variable is read from in `parse_dft_window` but is constant because it is never written to. This does not appear to be intended. This PR fixes that by removing the shadowing variable to let the parser write to the intended member variable instead of the local variable.

### DIFF
--- a/src/ipts/parser.cpp
+++ b/src/ipts/parser.cpp
@@ -137,57 +137,53 @@ void Parser::parse_reports(Reader reader)
 
 void Parser::parse_stylus_v1(Reader reader)
 {
-	StylusData stylus;
-
 	const auto stylus_report = reader.read<struct ipts_stylus_report>();
-	stylus.serial = stylus_report.serial;
+	this->stylus.serial = stylus_report.serial;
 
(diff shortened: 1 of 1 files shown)


StollD:
That variable is actually modified in the call to `this->on_dft`. The idea is that the DFT processing code stays independent from the code that interacts with the uinput device. That will make integrating it into the debugging tools easier, once I get around to doing that.

https://github.com/linux-surface/iptsd/blob/4613d19d56a11c2598d94e8c94e65df818344d9c/src/daemon/dft.cpp#L264-L265

A persistent variable is also neccessary, because not all information is sent in every DFT report. It is all a bit convoluted and sadly not well documented.

It also doesnt really matter that the variable in `parse_stylus_*` shadows the variable defined in the class, because `parse_stylus_*` and `parse_dft_window` are mutually exclusive, a device never uses both. But its still a sensible change, so thank you for it!

StollD (approved):

PR #114, opened 2023-02-24
Add IPTSD_CONFIG_FILE environment variable

danielzgtg:
Add an environment variable named `IPTSD_CONFIG_FILE` to load a configuration file from a custom location. This will mainly be used during development.

Currently the configuration is loaded from somewhere in `/etc/**` and `/usr/**`. Those directories are normally only writable by root. I have a hack where I did `sudo ln -s /home/home/Downloads/iptsd.conf /etc/iptsd.conf`. The situation is even worse for those developing on-device. With this PR, I hope to remove this hack.

This environment variable could also be used for automated testing.

### DIFF
--- a/src/config/config.cpp
+++ b/src/config/config.cpp
@@ -200,10 +200,20 @@ Config::Config(i16 vendor, i16 product, std::optional<const ipts::Metadata> meta
 	this->load_dir(IPTSD_PRESET_DIR, true);
 	this->load_dir("./etc/presets", true);
 
-	if (std::filesystem::exists(IPTSD_CONFIG_FILE))
-		ini_parse(IPTSD_CONFIG_FILE, parse_conf, this);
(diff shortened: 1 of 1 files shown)


src/config/config.cpp
```diff
@@ -197,13 +197,25 @@ Config::Config(i16 vendor, i16 product, std::optional<const ipts::Metadata> meta
 		this->invert_y = metadata->transform.yy < 0;
 	}
 
-	this->load_dir(IPTSD_PRESET_DIR, true);
-	this->load_dir("./etc/presets", true);
```
StollD, line 201: The presets must be always loaded. Otherwise iptsd won't know the screen size and inverted axes for the older devices (everything before SP7), and will not run or process data dumps from these devices.

StollD (changes requested):

StollD (approved):

StollD:
Thank you!

PR #131, opened 2023-05-17
fix: avoid unexpected failure in `override_dh_auto_install`

iamhyc:
`find ...` should always return at most one result located at the top-level of current folder.

The current behavior may encounter with multiple returns, and cause wrong usage of `meson install -C <WD>`.


StollD (approved):

StollD:
Thank you!

PR #136, opened 2023-06-16
contacts: detection: Take advantage of initializer lists if using Eigen 3.4

hmtheboy154:
Eigen has `EIGEN_VERSION_AT_LEAST` marco on Macros.h, with that we can keep the initializer lists code instead of removing it.


StollD:
Since there is no functional difference, and Eigen is backwards compatible, this isn't really required, sorry. I try to avoid using the preprocessor, and the code for old Eigen actually looks better in my opinon, because it fits into one line. 

Also, it doesnt really make sense to manually initialize an identity matrix instead of using `Matrix::Identity()`, so that was just me being dumb.

StollD:
Nevertheless, thank you.

PR #150, opened 2023-12-23
apps: visualization: Handle window events properly

iwanders:
Quick drive by PR to prevent the windows from being marked as unresponsive. This adds an event handler in the on data callback. I also tried setting the hints from [here](https://wiki.libsdl.org/SDL2/SDL_HINT_VIDEO_X11_NET_WM_PING), but I'm on Wayland and those hints don't work. This works reliably.

So without this, I get this window after 5 seconds (Gnome 45, Wayload, NixOS 23.11).

Screenshot from 2023-12-23 15-24-55

### DIFF
--- a/src/apps/visualization/visualize-sdl.hpp
+++ b/src/apps/visualization/visualize-sdl.hpp
@@ -68,6 +68,10 @@ public:
 	{
 		Visualize::on_data(data);
 
+		// Handle window events, such as the X11_NET_WM_PING
+		// event that is used to detect stuck programs.
+		SDL_PumpEvents();
+
 		constexpr usize FPS = 60;
 
 		// Limit how many times per seconds the screen is redrawn.


iwanders:
(force pushed to move the additions to be above the FPS variables instead of in between >_<)

src/apps/visualization/visualize-sdl.hpp
```diff
@@ -68,6 +68,13 @@ class VisualizeSDL : public Visualize {
 	{
 		Visualize::on_data(data);
 
+		// Handle window events, such as the X11_NET_WM_PING
+		// event that is used to detect stuck programs.
+		SDL_Event e;
+		while (SDL_PollEvent(&e))
```
StollD, line 74: Would `SDL_PumpEvents` work here too? It would avoid the need for an unused variable and an empty while loop.

If not, and the while loop needs to be kept, please make sure it is formatted according to the clang-format configuration. I let the CI check for that to keep everything consistent. In this case it would mean 

```cpp
while (...) {
}
```

even if that is uglier. You could consider moving the comment into the while loop to make it look nicer, but thats not a requirement for my side.

StollD (changes requested):

src/apps/visualization/visualize-sdl.hpp
iwanders: Good idea, I've confirmed that `SDL_PumpEvents` also works and pushed an improved commit.

StollD (approved):

StollD:
Thank you!

PR #151, opened 2023-12-25
Only use DFT button signal if it is from same group as position signal

quo:
Fixes https://github.com/linux-surface/linux-surface/issues/1318

### DIFF
--- a/src/ipts/parser.hpp
+++ b/src/ipts/parser.hpp
@@ -39,6 +39,7 @@ public:
 private:
 	struct ipts_dimensions m_dim {};
 	struct ipts_timestamp m_time {};
+	struct ipts_pen_metadata m_pen_meta {};
 
 public:
 	/*!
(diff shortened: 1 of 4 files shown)


src/ipts/parser.hpp
```diff
@@ -426,6 +430,12 @@ class Parser {
 		dft.rows = window.num_rows;
 		dft.type = window.data_type;
 
+		if (window.seq_num == m_pen_meta.seq_num &&
+			window.data_type == m_pen_meta.data_type) {
+			const auto g = m_pen_meta.group_counter;
```
StollD, line 435: Is there a reason for the temporary variable here? Why not assign to `dft.group` directly?

src/ipts/parser.hpp
```diff
@@ -426,6 +430,12 @@ class Parser {
 		dft.rows = window.num_rows;
 		dft.type = window.data_type;
 
+		if (window.seq_num == m_pen_meta.seq_num &&
+			window.data_type == m_pen_meta.data_type) {
```
StollD, line 434: Please format this according to the clang-format configuration, e.g.

```cpp
if (window.seq_num == m_pen_meta.seq_num &&
    window.data_type == m_pen_meta.data_type) {
```

StollD (changes requested):

src/ipts/parser.hpp
quo: GCC complains because the assignment to std::optional takes a reference, and taking a reference to a packed field is not allowed. Using an intermediate variable seemed like the simplest workaround (but I'm not a C++ expert, maybe there's a better way).

StollD (approved):

StollD:
Thank you!

Issue #102, opened 2022-12-30
Side button random presses Slim pen 2


Haislich:
While using the pen sometimes the input is misinterpreted, ad detects the press of the side button.
The solution mentioned  [here](https://github.com/quo/iptsd/issues/5#:~:text=edited-,setting%20ipts_dft_button_min_mag%20%3D%204000%20worked%20for%20me%20on%20the%20surface%20pro%20x.,-Author) seem to work partially but solved the problem for the most part.

https://user-images.githubusercontent.com/42048782/210048417-f4104b5a-3874-4e07-946f-fed07288ac17.mp4

https://user-images.githubusercontent.com/42048782/210048421-a866fe69-4dc8-48b7-ae4b-f8a32a2df871.mp4

StollD:
I am assuming that the first video is the behaviour with the proposed solution and the second one is the one without? Posting two videos without any kind of explanation isnt really helpful.

My guess is that the issue is fundamentally the same as #99: The values used in the DFT detection are (educated) guesses, and they might work better on one device than on another. The DFT code also doesn't seem to take noise into account (if that is even a possible issue). We would probably need a calibration tool similar to the one I made for the touchscreen so that people can fine tune the values.

Unfortunately, I can't really offer any helpful advice, since I don't own a device with a DFT stylus, and I am not really familiar with the theory behind that code either.

Haislich:
I'm sorry, my bad.
Both videos are before the suggested solution.
After the solution the clicks are so sporadic that are hard to catch on video.
Anyway the fine tuning tool might be a great idea, because for example palm rejection is still an hit or miss, the ability to enable/disable touch when the pen is touching the screen might be useful.

makinbacon21:
Note that the old suggested soln required a recomp, and that stuff is no longer there. I was able to resolve this using the newer version of the soln--edit `/etc/iptsd.conf` and set ButtonMinMag to 3000. You can tweak that value from there. Palm rejection I used the calibration tool for, same as touchpad calibration on SLS.

Issue #99, opened 2022-11-28
Surface Slim Pen 2 on SLS have mis-behaviour on soft pressure of eraser when pen is hover/touch toward to screen


NP-chaonay:
### Let talking little bit of history.

This come from past issue: https://github.com/quo/iptsd/issues/5

I have archived (I mean freezed) the above issue due to transition from quo's iptsd to official iptsd.

And found later that some of  problem mentioned there  still occured on official iptsd. 

so I decided to make new issue here.

BTW, I got suggestion form Dorain Stoll that: I should try [this comment](https://github.com/quo/iptsd/issues/5#issuecomment-1159294604) to solve DFT calib. which may cause the problem.

but as I remember, it dont work (see the test result I partially have noted there) but for sure, I will fully re-test it.

### ok let talking about problem it self below

it seem that eraser pressure sensors trigger BTN_STYLUS (this event I think is not about eraser, so it should not be triggered when eraser is not point toward the touch screen) even eraser is not point toward screen.

that happens on my SLS Ubuntu 22.04 (ithc and linux-surface iptsd)

I have made issue about it but closed because of transition to official iptsd from quo's iptsd

for my testing on my device

BTN_STYLUS active when
BTN_STYLUS active only when pen is point toward to screen AND in hover range AND eraser pressure is pressed.

And when hover in UI, it result in right click (according to mouse gnome-control-center) (some apps trigger what trigger by middle click, such as on link in google chrome)

NP-chaonay:
This comment is for test result of test "DFT Calib suggestion by Dorian Stoll"

if tested have finished/cancelled in either passed/failed way, it will be reported in new comment.

Status: not yet doing...

NP-chaonay:
> BTW, I got suggestion form Dorain Stoll that: I should try https://github.com/quo/iptsd/issues/5#issuecomment-1159294604 to solve DFT calib. which may cause the problem.

for that I willl post the result once I have test, so stay tuned (I have to do acedemic degree, so maybe using long time for it btw)

Haislich:
I'm having the issue, the side button acts like if it's pressed even if it is not the case.
Following the suggestions in the post setting  `dft_button_min_mag = 4000` in `iptsd/src/config/config.hpp` seems to have fixed.

makinbacon21:
[...]

PR #156, opened 2024-02-26
MPP v2+, Slim Pen 2, Surface Pro 9 improvements

iwanders:
Hi, this PR should address the long standing Slim Pen 2 button glitching from #102, #99 and https://github.com/quo/iptsd/issues/5.

Unfortunately I spent a lot of time at first trying to fix this by just using the current data and improving the detection / threshold. It all became easier when I took a step back and took a fresh look at all the data we get from the hardware with the MPP v2+ pens, without assuming `DftButton` actually holds button data. My analysis can be found at [iwanders/analysis-ithc-iptsd](https://github.com/iwanders/analysis-ithc-iptsd), that page should probably be linked in one of the reverse engineering issues, but I don't know which one would be preferred. Fyi @quo, @qzed, @StollD, huge shoutout to the work you have been doing thus far.

In this PR:

- Barrel button detection using the first `0x0a` window, using row 4 and 5. Implementation is a bit hacky as the current data isn't grouped, but I saw @StollD  modifying the parser handling this week, so filing this PR may help shape the direction to work towards.
- Contact detection using the (previously called) `Position2` frame, row 2 and 3.
- Add config flag to specify Microsoft Pen Protocol 2.0+.
- Add a preset for the Surface Pro 9.
- Add some logging around the configuration loading, I derped around a bit only to discover my config in `/etc/iptsd/iptsd.conf` wasn't being loaded and I ended up with the defaults from the code, oops, there should now be a warning if no configuration is loaded. Should've probably been a different PR, but I ran into it while working on this.
- Logic is based on recordings (both Windows & Linux) on a SP9 with Slim Pen 2, Metapen M2, Pen from AliExpress, and against data provided by @NP-chaonay in one of the issues.
- I tried to make the commits in the appropriate style, but it may be easier to just squash them, I don't think the individual commits pass the linter.

Known limitations / deficiencies:
- The button detection a approach means that I lose the eraser functionality for the Pen from AliExpress (eraser is done with a button, which also triggers the `0x0a` row 4 & 5 switch), but I don't know how to fix that without using multiple frames instead of a single frame & window, we'd have to detect the alternating pattern.
- This PR does not address the situation where the position interpolation [fails here](https://github.com/linux-surface/iptsd/blob/f59b72646afd931e5a1013e6c76548d750979229/src/core/generic/dft.hpp#L107) and results in the pen being lifted.
- Have not done extensive prolonged testing of this approach. I cannot test on different hardware. At some point this repo should have integration tests running against data dumps.

Filing this as a draft PR for discussion, current window naming is obviously a stop gap solution, hopefully this PR also gives others a chance to help test, I'll have limited time to test, but will (obviously) make time to incorporate feedback and answer questions.

### DIFF
--- a/etc/presets/surface-pro-9.conf
+++ b/etc/presets/surface-pro-9.conf
@@ -1,7 +1,11 @@
+[Device]
+Vendor = 0x045E
+Product = 0x0C52
+
 [Stylus]
 MPPVersion = v2
 
 [DFT]
-# This is really high, but SP2 creates a strong signal.
 ButtonMinMag = 50000
+ContactMinMag = 50000
 PositionMinMag = 500
--- a/etc/presets/surface-pro-9.conf
+++ b/etc/presets/surface-pro-9.conf
@@ -1,9 +1,0 @@
-[Device]
-Vendor = 0x045E
-Product = 0x0C52
-
-[DFT]
-ButtonMinMag = 10000
-Mpp2ContactMinMag = 50000
-Mpp2ButtonMinMag = 50000
-PositionMinMag = 500
--- a/etc/presets/surface-pro-9.conf
+++ b/etc/presets/surface-pro-9.conf
@@ -2,5 +2,6 @@
 MPPVersion = v2
 
 [DFT]
-ButtonMinMag = 1000
+# This is really high, but SP2 creates a strong signal.
+ButtonMinMag = 50000
 PositionMinMag = 500
--- a/etc/presets/surface-pro-9.conf
+++ b/etc/presets/surface-pro-9.conf
@@ -2,10 +2,8 @@
 Vendor = 0x045E
 Product = 0x0C52
 
-[Stylus]
-MPPVersion = v2
-
 [DFT]
-ButtonMinMag = 50000
-ContactMinMag = 50000
+ButtonMinMag = 10000
+Mpp2ContactMinMag = 50000
+Mpp2ButtonMinMag = 50000
 PositionMinMag = 500
--- a/etc/presets/surface-pro-9.conf
+++ b/etc/presets/surface-pro-9.conf
@@ -1,0 +1,6 @@
+[Stylus]
+MPPVersion = v2
+
+[DFT]
+ButtonMinMag = 1000
+PositionMinMag = 500
--- a/src/core/generic/config.hpp
+++ b/src/core/generic/config.hpp
@@ -57,7 +57,6 @@ public:
 	usize dft_button_min_mag = 1000;
 	usize dft_freq_min_mag = 10000;
 	usize dft_tilt_min_mag = 10000;
-	usize dft_contact_min_mag = 10000;
 	usize dft_mpp2_button_min_mag = 50000;
 	usize dft_mpp2_contact_min_mag = 50000;
 	f64 dft_tilt_distance = 0.6;
--- a/src/core/generic/config.hpp
+++ b/src/core/generic/config.hpp
@@ -17,14 +17,6 @@ namespace iptsd::core {
 
 class Config {
 public:
-	// Enum to denote which Pen Protocol is being used.
-	enum class MPPVersion {
-		// Used for Microsoft Pen Protocol v1 compliant pens.
-		V1,
-		// Used for Microsoft Pen Protocol v2 compliant pens.
-		V2,
-	};
-
 	// [Config]
 	bool invert_x = false;
 	bool invert_y = false;
@@ -57,7 +49,6 @@ public:
 	// [Stylus]
 	bool stylus_disable = false;
 	f64 stylus_tip_distance = 0;
-	MPPVersion stylus_mpp_version = MPPVersion::V1;
 
 	// [DFT]
 	usize dft_position_min_amp = 50;
@@ -67,6 +58,8 @@ public:
 	usize dft_freq_min_mag = 10000;
 	usize dft_tilt_min_mag = 10000;
 	usize dft_contact_min_mag = 10000;
+	usize dft_mpp2_button_min_mag = 50000;
+	usize dft_mpp2_contact_min_mag = 50000;
 	f64 dft_tilt_distance = 0.6;
 
 public:
--- a/src/core/generic/config.hpp
+++ b/src/core/generic/config.hpp
@@ -66,6 +66,7 @@ public:
 	usize dft_button_min_mag = 1000;
 	usize dft_freq_min_mag = 10000;
 	usize dft_tilt_min_mag = 10000;
+	usize dft_contact_min_mag = 10000;
 	f64 dft_tilt_distance = 0.6;
 
 public:
--- a/src/core/generic/config.hpp
+++ b/src/core/generic/config.hpp
@@ -17,6 +17,14 @@ namespace iptsd::core {
 
 class Config {
 public:
+	// Enum to denote which Pen Protocol is being used.
+	enum class MPPVersion {
+		// Used for Microsoft Pen Protocol v1 compliant pens.
+		V1,
+		// Used for Microsoft Pen Protocol v2 compliant pens.
+		V2,
+	};
+
 	// [Config]
 	bool invert_x = false;
 	bool invert_y = false;
@@ -49,6 +57,7 @@ public:
 	// [Stylus]
 	bool stylus_disable = false;
 	f64 stylus_tip_distance = 0;
+	MPPVersion mpp_version = MPPVersion::V1;
 
 	// [DFT]
 	usize dft_position_min_amp = 50;
--- a/src/core/generic/config.hpp
+++ b/src/core/generic/config.hpp
@@ -57,7 +57,7 @@ public:
 	// [Stylus]
 	bool stylus_disable = false;
 	f64 stylus_tip_distance = 0;
-	MPPVersion mpp_version = MPPVersion::V1;
+	MPPVersion stylus_mpp_version = MPPVersion::V1;
 
 	// [DFT]
 	usize dft_position_min_amp = 50;
--- a/src/core/generic/dft.hpp
+++ b/src/core/generic/dft.hpp
@@ -169,11 +169,6 @@ private:
 	 */
 	void handle_button(const ipts::DftWindow &dft)
 	{
-		// Only use this for pen v1.
-		if (m_config.mpp_version != Config::MPPVersion::V1) {
-			return;
-		}
-
 		if (dft.x.empty())
 			return;
 
@@ -199,7 +194,10 @@ private:
 			rubber = val > 0;
 		}
 
-		m_stylus.button = button;
+		// Only set the button value if we're using a v1 pen.
+		if (m_config.mpp_version == Config::MPPVersion::V1) {
+			m_stylus.button = button;
+		}
 		m_stylus.rubber = rubber;
 	}
 
@@ -227,7 +225,9 @@ private:
 
 
 	/*!
-	 * Determines the current button state from the 0x0a frame.
+	 * Determines the current button state from the 0x0a frame, it can
+	 * only be used for MPP v2 pens. The eraser is still obtained from the
+     * phase using the button frame.
 	 */
 	void handle_dft_0x0a(const ipts::DftWindow &dft)
 	{
@@ -250,16 +250,12 @@ private:
 		if (mag_4 < threshold && mag_5 < threshold) {
 			// Not enough signal, lets disable the button
 			m_stylus.button = false;
-			m_stylus.rubber = false;
 			return;
 		}
 
 		// One of them is above the threshold, if 5 is higher than 4, button
 		// is held.
 		m_stylus.button = mag_4 < mag_5;
-
-		// This needs a todo still :)
-		m_stylus.rubber = false;
 	}
 
 	/*!
--- a/src/core/generic/dft.hpp
+++ b/src/core/generic/dft.hpp
@@ -34,6 +34,7 @@ private:
 	// allows comparing against this group and only using the first 0x0a
 	// frame.
 	std::optional<u32> m_dft_0x0a_group = std::nullopt;
+
 public:
 	DftStylus(Config config, const std::optional<const ipts::Metadata> &metadata)
 		: m_config {std::move(config)},
@@ -223,11 +224,10 @@ private:
 		}
 	}
 
-
 	/*!
 	 * Determines the current button state from the 0x0a frame, it can
 	 * only be used for MPP v2 pens. The eraser is still obtained from the
-     * phase using the button frame.
+	 * phase using the button frame.
 	 */
 	void handle_dft_0x0a(const ipts::DftWindow &dft)
 	{
--- a/src/core/generic/dft.hpp
+++ b/src/core/generic/dft.hpp
@@ -29,11 +29,10 @@ private:
 	std::optional<u32> m_group = std::nullopt;
 
 	// This is a bit of a hack, for the MPP v2 button detection we only
-	// care about the first 0x0a dft window, but there's two in the frame,
-	// here we keep track of the group when 0x0a was encountered, this
-	// allows comparing against this group and only using the first 0x0a
-	// frame.
-	std::optional<u32> m_dft_0x0a_group = std::nullopt;
+	// care about the first binary (0x0a) dft window, but there's two in the
+	// frame, here we keep track of the group when 0x0a was encountered,
+	// this allows comparing against this group and only using the first frame.
+	std::optional<u32> m_mppv2_binary_group = std::nullopt;
 
 	// Boolean to track whether the pen is in contact according to mpp v2.
 	// This is used to override the contact state in the pressure frame handling.
@@ -232,8 +231,7 @@ private:
 			m_stylus.contact = true;
 			m_stylus.pressure = std::clamp(p, 0.0, 1.0);
 		} else {
-			m_stylus.contact =
-				m_mppv2_in_contact.has_value() ? m_mppv2_in_contact.value() : false;
+			m_stylus.contact = m_mppv2_in_contact.value_or(false);
 			m_stylus.pressure = 0;
 		}
 	}
@@ -245,24 +243,23 @@ private:
 	 */
 	void handle_dft_binary_mpp_2(const ipts::DftWindow &dft)
 	{
-		// Clearing the state in case we can't determine it.
-		m_mppv2_button = std::nullopt;
-
 		if (dft.x.size() <= 5) { // not sure if this can happen?
 			return;
 		}
 
-		// Second time we see the 0x0a frame in this group, skip it.
-		if (!dft.group.has_value() || m_dft_0x0a_group == dft.group)
+		// Second time we see this dft window in this group, skip it.
+		if (!dft.group.has_value() || m_mppv2_binary_group == dft.group)
 			return;
+		m_mppv2_binary_group = dft.group;
 
-		m_dft_0x0a_group = dft.group;
+		// Clearing the state in case we can't determine it.
+		m_mppv2_button = std::nullopt;
 
 		// Now, we can process the frame to determine button state.
 		// First, collapse x and y, they convey the same information.
 		const auto mag_4 = dft.x[4].magnitude + dft.y[4].magnitude;
 		const auto mag_5 = dft.x[5].magnitude + dft.y[5].magnitude;
-		const auto threshold = 2 * m_config.dft_button_min_mag;
+		const auto threshold = 2 * m_config.dft_mpp2_button_min_mag;
 
 		if (mag_4 < threshold && mag_5 < threshold) {
 			// Not enough signal to make a decision.
--- a/src/core/generic/dft.hpp
+++ b/src/core/generic/dft.hpp
@@ -35,6 +35,10 @@ private:
 	// frame.
 	std::optional<u32> m_dft_0x0a_group = std::nullopt;
 
+	// Boolean to track whether the pen is in contact, used by the method
+	// that processes the pressure frame.
+	bool m_mppv2_in_contact {false};
+
 public:
 	DftStylus(Config config, const std::optional<const ipts::Metadata> &metadata)
 		: m_config {std::move(config)},
@@ -56,6 +60,9 @@ public:
 			break;
 		case ipts::protocol::dft::Type::Pressure:
 			this->handle_pressure(dft);
+			break;
+		case ipts::protocol::dft::Type::Position2:
+			this->handle_position2(dft);
 			break;
 		case ipts::protocol::dft::Type::Dft0x0a:
 			this->handle_dft_0x0a(dft);
@@ -219,7 +226,7 @@ private:
 			m_stylus.contact = true;
 			m_stylus.pressure = std::clamp(p, 0.0, 1.0);
 		} else {
-			m_stylus.contact = false;
+			m_stylus.contact = false || m_mppv2_in_contact; // NOLINT
 			m_stylus.pressure = 0;
 		}
 	}
@@ -232,6 +239,11 @@ private:
 	void handle_dft_0x0a(const ipts::DftWindow &dft)
 	{
 		if (m_config.mpp_version != Config::MPPVersion::V2) {
+			return;
+		}
+
+		if (dft.x.size() <= 5) { // not sure if this can happen?
+			m_stylus.button = false;
 			return;
 		}
 
@@ -256,6 +268,35 @@ private:
 		// One of them is above the threshold, if 5 is higher than 4, button
 		// is held.
 		m_stylus.button = mag_4 < mag_5;
+	}
+
+	/*!
+	 * Determines whether the pen is making contact with the screen, it can
+	 * only be used for MPP v2 pens.
+	 */
+	void handle_position2(const ipts::DftWindow &dft)
+	{
+		if (m_config.mpp_version != Config::MPPVersion::V2) {
+			return;
+		}
+
+		// Set to false in case we return early.
+		m_mppv2_in_contact = false;
+
+		if (dft.x.size() <= 3) { // not sure if this can happen?
+			return;
+		}
+
+		const auto mag_2 = dft.x[2].magnitude + dft.y[2].magnitude;
+		const auto mag_3 = dft.x[3].magnitude + dft.y[3].magnitude;
+
+		const auto threshold = 2 * m_config.dft_contact_min_mag;
+		if (mag_2 < threshold && mag_3 < threshold) {
+			return;
+		}
+
+		// The pen switches the row from two to three when there's contact.
+		m_mppv2_in_contact = mag_2 < mag_3;
 	}
 
 	/*!
--- a/src/core/generic/dft.hpp
+++ b/src/core/generic/dft.hpp
@@ -35,9 +35,13 @@ private:
 	// frame.
 	std::optional<u32> m_dft_0x0a_group = std::nullopt;
 
-	// Boolean to track whether the pen is in contact, used by the method
-	// that processes the pressure frame.
-	bool m_mppv2_in_contact {false};
+	// Boolean to track whether the pen is in contact according to mpp v2.
+	// This is used to override the contact state in the pressure frame handling.
+	std::optional<bool> m_mppv2_in_contact = std::nullopt;
+
+	// Boolean to track whether the button is held according to mpp v2.
+	// This is used to override the button state in the button frame handling.
+	std::optional<bool> m_mppv2_button = std::nullopt;
 
 public:
 	DftStylus(Config config, const std::optional<const ipts::Metadata> &metadata)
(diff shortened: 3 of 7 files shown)


src/core/generic/config.hpp
```diff
@@ -49,6 +57,7 @@ class Config {
 	// [Stylus]
 	bool stylus_disable = false;
 	f64 stylus_tip_distance = 0;
+	MPPVersion stylus_mpp_version = MPPVersion::V1;
```
iwanders, line 60: I grouped this under stylus, even though it's only really used in the `dft` files... but it still seemed like the most appropriate spot.

src/core/generic/dft.hpp
```diff
@@ -28,6 +28,17 @@ class DftStylus {
 	i32 m_imag = 0;
 	std::optional<u32> m_group = std::nullopt;
 
+	// This is a bit of a hack, for the MPP v2 button detection we only
+	// care about the first 0x0a dft window, but there's two in the frame,
+	// here we keep track of the group when 0x0a was encountered, this
+	// allows comparing against this group and only using the first 0x0a
+	// frame.
+	std::optional<u32> m_dft_0x0a_group = std::nullopt;
```
iwanders, line 36: Like the comment says, this is a hack, just like `m_mppv2_in_contact` actually. In the ideal situation we would not have to track this state between individual `input()` calls, but get some data structure that contains everything from the group and then can decided on the current stylus state using the full state of the frame.

src/core/generic/dft.hpp
```diff
@@ -206,11 +226,79 @@ class DftStylus {
 			m_stylus.contact = true;
 			m_stylus.pressure = std::clamp(p, 0.0, 1.0);
 		} else {
-			m_stylus.contact = false;
+			m_stylus.contact = false || m_mppv2_in_contact; // NOLINT
```
iwanders, line 229: Added a nolint here, clang tidy didn't like the unnecessary `false ||`, but think it is good to have it for the symmetry, it is false, unless the heuristic for mppv2 decides it is true. Don't have strong feelings though, but I disagreed with clang tidy and added this.

src/core/generic/dft.hpp
StollD: Tracking state is fine. We either need to track state here or in the parser if I am not mistaken, because the information is in different report frames. Here is better.

src/core/generic/dft.hpp
```diff
@@ -28,6 +28,17 @@ class DftStylus {
...
+	// This is a bit of a hack, for the MPP v2 button detection we only
+	// care about the first 0x0a dft window, but there's two in the frame,
+	// here we keep track of the group when 0x0a was encountered, this
+	// allows comparing against this group and only using the first 0x0a
+	// frame.
+	std::optional<u32> m_dft_0x0a_group = std::nullopt;
+
+	// Boolean to track whether the pen is in contact, used by the method
+	// that processes the pressure frame.
+	bool m_mppv2_in_contact {false};
```
StollD, line 40: Nitpick:

```suggestion
	bool m_mppv2_in_contact = false;
```

src/core/generic/dft.hpp
StollD: I don't really mind this, but please add a comment explaining the reasoning behind it, because otherwise I will just end up removing it in a few months xD

src/core/generic/dft.hpp
```diff
@@ -185,7 +202,10 @@ class DftStylus {
 			rubber = val > 0;
 		}
 
-		m_stylus.button = button;
+		// Only set the button value if we're using a v1 pen.
+		if (m_config.stylus_mpp_version == Config::MPPVersion::V1) {
+			m_stylus.button = button;
+		}
```
StollD, line 208: Nitpick:

```suggestion
		if (m_config.stylus_mpp_version == Config::MPPVersion::V1)
			m_stylus.button = button;
```

src/core/linux/config-loader.hpp
```diff
@@ -48,14 +50,21 @@ class ConfigLoader {
 		 * known working main system configuration.
 		 */
 		if (const char *config_file_path = std::getenv("IPTSD_CONFIG_FILE")) {
+			spdlog::info("Loading config {}, specified with IPTSD_CONFIG_FILE.",
```
StollD, line 53: I like the idea, but maybe replace all of these log statements with a single one in `load_file`?

src/core/linux/config-loader.hpp
```diff
@@ -48,14 +50,21 @@ class ConfigLoader {
...
-		if (std::filesystem::exists(common::buildopts::ConfigFile))
+		if (std::filesystem::exists(common::buildopts::ConfigFile)) {
+			spdlog::info("Loading config {}.", common::buildopts::ConfigFile);
 			this->load_file(common::buildopts::ConfigFile);
+		}
 
 		this->load_dir(common::buildopts::ConfigDir, false);
+
+		if (!m_loaded_config)
+			spdlog::warn("No config file loaded at all, this is not good.");
```
StollD, line 67: This is not really an issue, because all config values have defaults, and newer devices don't need presets, so I wouldn't print a warning because of it.

src/core/linux/config-loader.hpp
```diff
@@ -199,7 +212,21 @@ class ConfigLoader {
...
+					value = Config::MPPVersion::V1;
+				} else if (mpp_version == "v2") {
+					value = Config::MPPVersion::V2;
+				} else {
+					throw common::Error<Error::ParsingInvalidValue> {
+						"Stylus mpp_version was not 'v1' or 'v2', got: " +
+						mpp_version};
+				}
+			}
+		} else
```
StollD, line 229: How about just treating this as a number?

Alternatively, I would prefer if this was parsed into a temporary string and then mapped in `load_file`.

src/ipts/parser.hpp
```diff
@@ -71,7 +71,6 @@ class Parser {
 	{
 		Reader reader(data);
 		reader.skip(header);
-
```
StollD, line 74: Nitpick: Please revert this.

src/ipts/protocol/dft.hpp
```diff
@@ -15,8 +15,10 @@ constexpr u8 PRESSURE_ROWS = 6;
 
 enum class Type : u8 {
 	Position = 6,
+	Position2 = 7,
 	Button = 9,
 	Pressure = 11,
+	Dft0x0a = 0x0a,
```
StollD, line 21:
```suggestion
	Position = 0x6,
	Position2 = 0x7,
	Button = 0x9,
	Button2 = 0xA,
	Pressure = 0xB,
```

You could also consider putting the relation to MPP into the name, e.g. `PositionMPP_1` and `PositionMPP_2`.

src/core/generic/dft.hpp
```diff
@@ -206,11 +226,79 @@ class DftStylus {
...
-			m_stylus.contact = false;
+			m_stylus.contact = false || m_mppv2_in_contact; // NOLINT
 			m_stylus.pressure = 0;
 		}
 	}
 
+	/*!
+	 * Determines the current button state from the 0x0a frame, it can
+	 * only be used for MPP v2 pens. The eraser is still obtained from the
+	 * phase using the button frame.
```
StollD, line 237: From your notes I gather that this can't be used with MPP v1 pens because it contains garbage?

Is there any way to detect if the data inside is usable, and adjust the behaviour of the DFT parser accordingly? Dropping the need for a config option.

iwanders: > Is there any way to detect if the data inside is usable, and adjust the behaviour of the DFT parser accordingly? [...] I don't really like the idea of having to make the model / type of the pen a config option

Hmm, yeah, I'm not a fan of it either, because it means you can't use a mpp v1 pen without changing the config.

It may actually be possible... if a v1 pen is present, the magnitudes for the `0x0a` frame are very small (100s). So what we could do I guess is check if the signal in the `0x0a` frame is above a threshold (50000); if so, it trumps the button detection from the button frame.

Let me do a bit more exploration around this to see if I can make it work without explicit configuration.

iwanders:
>> but I don't know how to fix that without using multiple frames instead of a single frame & window, we'd have to detect the alternating pattern.

> If adding a temporal component to the DFT parser improves the situation, I'd be all for it.

Problem is that in the CN pen, the 4 to 5 switch still happens when the eraser button is pressed;

Image for depressing eraser, then make screen contact. :

2024_02_25_linux_cn_hover_erase_touch


For hover, press barrel, touch screen, release barrel, move, raise from contact:

2024_02_25_linux_cn_hover_button_click_touch_release_wait_hover

The only real distinction is the alternating signal, which is on like `0x0a 2&3  e&f` (both first and second window). We could detect that for this pen and 'solve it'.

The problem is that the SP pen on linux creates a binary code (probably a pen identifier, just like the M2 on Windows, see [this section](https://github.com/iwanders/analysis-ithc-iptsd?tab=readme-ov-file#barrel-test-sp-linux) and we don't want that binary code to trigger the eraser detection. We could do something ugly where we check if 6/6 frames were perfectly alternating or something, but I'm really worried that this binary code is like the 'marker color', a unique id for each pen to allow the screen to distinguish two of them and some pens may just have an id that creates a binary sequence that perfectly alternates. In short; I don't yet know how to solve this, only thing I can think of is making the eraser signal trump the button action... We could explore that, I don't think I've ever had an accidental eraser action, but the whole intent of this PR is to use the strong binary indicators available in v2+.

I had hoped the whole design in the CN pen with the whole eraser & button having almost the same signal because the emitter end is actually facing the screen was just a cost cutting measure, but according to [this document](https://learn.microsoft.com/en-us/windows-hardware/design/component-guidelines/windows-pen-designs#physical-design) the eraser in a side button is an approved design.

Funny thing is that [this section](https://learn.microsoft.com/en-us/windows-hardware/design/component-guidelines/windows-pen-states#windows-pen-state-transitions) then describes that the protocol can't actually handle that well:

> Unlike tail-end eraser implementations, button-based implementations can physically allow the user to activate/deactivate the erase affordance without the pen transitioning through the "out of range" state. However, this is not supported by the underlying protocol.

src/core/linux/config-loader.hpp
iwanders: It is implicit... which means that the ones that know it is implicit won't encounter the footgun. I wasted 30 minutes before I started the daemon with `strace` to see I placed my config file in the wrong location >_<

Either way, your code, your call; I'll remove it.

src/core/generic/dft.hpp
iwanders: Okay, I think this is possible, we can use a `std::optional<bool>` to specify whether the v2 signal has been detected. 

I do think it is a good idea to allow different minimum magnitudes for the v2 button vs the v1 button detection though, the scales are completely different;

```
SP, without touching button.
mag4 and 5: 3668922    69
Button mag x and y: 4321    25085
mag4 and 5: 3583622    54
Button mag x and y: 4930    24469
mag4 and 5: 3627867    34
Button mag x and y: 5128    25154

SP, 5mm, going into button
Button mag x and y: 4    16
mag4 and 5: 333657    2
Button mag x and y: 39464    33930
mag4 and 5: 26    333492
Button mag x and y: 41810    34442

SP, touching, going into button
Button mag x and y: 5    32
mag4 and 5: 3980021    61
Button mag x and y: 308425    521441
mag4 and 5: 114    3756423

m1, 5mm, going into button
Button mag x and y: 5    0
mag4 and 5: 20    15
Button mag x and y: 26065    26065
mag4 and 5: 8    10
Button mag x and y: 23953    22730

m1, 5mm
Button mag x and y: 17945    16769
mag4 and 5: 70    155
Button mag x and y: 19109    16840
mag4 and 5: 127    37
Button mag x and y: 19825    16160
mag4 and 5: 43    19

m1, touching
Button mag x and y: 434394    273589
mag4 and 5: 42    31
Button mag x and y: 558101    235801
mag4 and 5: 29    74

m2, touching
Button mag x and y: 275221    193257
mag4 and 5: 8502    2500942
Button mag x and y: 276692    186628
mag4 and 5: 8338    2490670

```

</detail>

If we use the same threshold, I can only detect my `m1` pen's button when it is super close to the screen.

src/core/generic/dft.hpp
iwanders: Changed this line in 772fbaf8003288bab37fe3a2875d2e796d0d2506, it is now a ternary that checks the optional value.

src/core/generic/dft.hpp
iwanders: Changed in 772fbaf8003288bab37fe3a2875d2e796d0d2506.

src/core/generic/dft.hpp
iwanders: Removed this check completely in 772fbaf8003288bab37fe3a2875d2e796d0d2506.

src/core/linux/config-loader.hpp
iwanders: Removed in 772fbaf8003288bab37fe3a2875d2e796d0d2506.

src/core/linux/config-loader.hpp
StollD: My point is kinda that if you *want* to `rm -rf` your config then iptsd shouldn't try to prevent you from doing it, because it won't break anything.

How about this? Neutrally printing that nothing was done seems like a good compromise.

```suggestion
			spdlog::info("No config file loaded, using default values.");
```

src/core/generic/dft.hpp
iwanders: Ok, this is now implemented under 772fbaf8003288bab37fe3a2875d2e796d0d2506.

I'll need to give it a bit more thorough testing, but from the data I see no reason why this shouldn't work. And brief tests with all four pens show that this works, it does allow me to just switch between the SP, M1, M2 and CN pens without any problems in regards to button detection.

Great idea to just detect this, I did add different magnitude detection for the v2 signals to the configuration, because they do have a very different amplitude from the v1 button signal, but that's still a lot better than having to set a boolean.

src/core/linux/config-loader.hpp
iwanders: Moved to a single print in the `load_file` method in 9aee2cd463207fb476c5296c90569653ba3aa7f9.

src/core/generic/dft.hpp
```diff
@@ -206,11 +232,74 @@ class DftStylus {
 			m_stylus.contact = true;
 			m_stylus.pressure = std::clamp(p, 0.0, 1.0);
 		} else {
-			m_stylus.contact = false;
+			m_stylus.contact =
+				m_mppv2_in_contact.has_value() ? m_mppv2_in_contact.value() : false;
```
StollD, line 236:
```suggestion
			m_stylus.contact = m_mppv2_in_contact.value_or(false);
```

src/core/generic/dft.hpp
iwanders: Unresolving this, need to change the variable name.

src/core/generic/dft.hpp
```diff
@@ -206,11 +232,74 @@ class DftStylus {
...
 
+	/*!
+	 * Determines the current button state from the 0x0a frame, it can
+	 * only be used for MPP v2 pens. The eraser is still obtained from the
+	 * phase using the button frame.
+	 */
+	void handle_dft_binary_mpp_2(const ipts::DftWindow &dft)
+	{
+		// Clearing the state in case we can't determine it.
+		m_mppv2_button = std::nullopt;
```
iwanders, line 249: This is a bug, it needs to be below the group check.

iwanders:
A few quick notes before work from my testing yesterday evening in comments.

Another limitation that currently affects mpp v2 pens is that the tilt is currently taken from Position, row 1, but that signal is only present when the stylus is touching the screen. I wonder if we should process `Position2` if present instead. That has a solid signal on the tilt. Or we can - when we process `Position2` for the contact detection store the dft values of the tilt row and consume that in the normal `Position` processing.

StollD:
> Another limitation that currently affects mpp v2 pens is that the tilt is currently taken from Position, row 1, but that signal is only present when the stylus is touching the screen.

Do you know how this is handled on Windows? On the "old" devices without a DFT stylus, the tilt is 0 when the pen tip is not touching the display.

etc/presets/surface-pro-9.conf
```diff
@@ -0,0 +1,9 @@
+[Device]
+Vendor = 0x045E
+Product = 0x0C52
+
+[DFT]
+ButtonMinMag = 10000
+Mpp2ContactMinMag = 50000
+Mpp2ButtonMinMag = 50000
+PositionMinMag = 500
```
StollD, line 9: Since the other values in the file are just defaults, is this change significant enough to warrant a preset file?

src/core/linux/config-loader.hpp
```diff
@@ -52,8 +52,9 @@ class ConfigLoader {
 			return;
 		}
 
-		if (std::filesystem::exists(common::buildopts::ConfigFile))
+		if (std::filesystem::exists(common::buildopts::ConfigFile)) {
 			this->load_file(common::buildopts::ConfigFile);
+		}
```
StollD, line 57: Please revert this change

src/core/linux/errors.hpp
```diff
@@ -9,6 +9,7 @@ namespace iptsd::core::linux {
 
 enum class Error {
 	ParsingFailed,
+	ParsingInvalidValue,
```
StollD, line 12: Since this value is used nowhere, this should be reverted

etc/presets/surface-pro-9.conf
iwanders: Good point, we can probably drop it now, let me explore that.

iwanders:
>>  Another limitation that currently affects mpp v2 pens is that the tilt is currently taken from Position, row 1, but that signal is only present when the stylus is touching the screen.
>
> Do you know how this is handled on Windows? On the "old" devices without a DFT stylus, the tilt is 0 when the pen tip is not touching the display.

I checked, I'm _really_ surprised, but it does indeed become zero if the stylus isn't touching the screen. Even Microsoft's DigiInfo reports zero tiltx and tilty if the stylus isn't in contact. I installed Krita on windows, with the input API `Windows 8+ Pointer Input (Windows Ink)` this is also the behaviour. Same with the [pointer events](https://patrickhlauke.github.io/touch/pen-tracker/) webapp. 

There's no technical reason though... Like we do have the data in `Position2` row `1` when the pen is raised, did a quick hack with just doing:
```cpp
		case ipts::protocol::dft::Type::PositionMPP_2:
			this->handle_position_mpp_2(dft);
			this->handle_position(dft);
			break;
```

And in that case I do end up with a tilt that updates while the pen is raised, both in Krita as in the pointer events webapp. I'm astonished Windows doesn't do this.

Also, speaking about dft updates... we currently call `on_stylus` every dft window:
https://github.com/linux-surface/iptsd/blob/f59b72646afd931e5a1013e6c76548d750979229/src/core/generic/application.hpp#L232

Should we think about a way where we can ensure that we only send that once for each group? Such that we don't have to worry about keeping the stylus data in a consistent state as we process the various dft windows in one group? Like, currently we consume the previous' cycle `0x0a` DFT window on the next group's `Button` window. It's not a huge concern imo, but something to think about.

StollD:
Tilt is calculated from two position signals, right? How does this work when one of the transmitters is out of range? Maybe thats why they only use the data when it is touching the display: To make sure both transmitters are in proximity.

Or this is just some weirdness that they inherited from N-Trig and never bothered to change.

I think by default iptsd should follow the Windows behaviour, but this sounds like a useful thing to put behind a config option. Although maybe that should be a seperate PR?

> Should we think about a way where we can ensure that we only send that once for each group? Such that we don't have to worry about keeping the stylus data in a consistent state as we process the various dft windows in one group? Like, currently we consume the previous' cycle `0x0a` DFT window on the next group's `Button` window. It's not a huge concern imo, but something to think about.

Another thing I noticed on my SB2 is that it only updates the tilt if I move the stylus. It sounds stupid with how precise these coordinates are supposed to be (1/9600 of the screen), but I can carefully tilt the stylus without the value changing.

Maybe that is something that the DFT parser could emulate.

etc/presets/surface-pro-9.conf
iwanders: You get a bit more range, but it's a pretty minimal change. Not worth the overhead in this repo I think; removed in b4ef3bdd6243c21f4af33c86abb84ff6d74a5e17.

src/core/linux/config-loader.hpp
iwanders: Reverted as part of 1623671b32151f52cefffb0043235bb0ef515813.

src/core/linux/errors.hpp
iwanders: Removed in 5f399fee731786d990e3e2982455efa4bf299621.

src/core/generic/dft.hpp
```diff
@@ -206,11 +232,74 @@ class DftStylus {
...
+		if (!dft.group.has_value() || m_dft_0x0a_group == dft.group)
+			return;
+
+		m_dft_0x0a_group = dft.group;
+
+		// Now, we can process the frame to determine button state.
+		// First, collapse x and y, they convey the same information.
+		const auto mag_4 = dft.x[4].magnitude + dft.y[4].magnitude;
+		const auto mag_5 = dft.x[5].magnitude + dft.y[5].magnitude;
+		const auto threshold = 2 * m_config.dft_button_min_mag;
```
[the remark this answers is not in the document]
iwanders: Addressed in 5f399fee731786d990e3e2982455efa4bf299621.

src/core/generic/dft.hpp
iwanders: Addressed in 5f399fee731786d990e3e2982455efa4bf299621.

src/core/generic/dft.hpp
iwanders: Incorporated in 5f399fee731786d990e3e2982455efa4bf299621.

src/core/linux/config-loader.hpp
iwanders: Added back in 1623671b32151f52cefffb0043235bb0ef515813, please resolve this if you're happy with the current form.

src/core/generic/dft.hpp
iwanders: Changed to `m_mppv2_binary_group` in 5f399fee731786d990e3e2982455efa4bf299621.

iwanders:
> Tilt is calculated from two position signals, right? How does this work when one of the transmitters is out of range? Maybe thats why they only use the data when it is touching the display: To make sure both transmitters are in proximity.

If either signal strength goes below a threshold we just lose the pointer position and/or tilt signal. Thing is, in Krita you can see the outline of the cursor update to show the tilt and size of the brush before you start drawing, so there's a lot of value in updating the tilt before you touch the screen imo.

> I think by default iptsd should follow the Windows behaviour, but this sounds like a useful thing to put behind a config option. Although maybe that should be a seperate PR?

Yeah, we can do that, only real outstanding items with this piece of work:
- a bit more testing, ideally with some others with Slim Pen2 & SP7+, but it may be easier to consume for others after it is merged.
- The lack of eraser functionality with my Chinese pen, but the only solution I see there is being able to let the eraser disable the button state. Thoughts on this?

> It sounds stupid with how precise these coordinates are supposed to be (1/9600 of the screen)

I expect that's part of the filtering done by the driver / hardware... I wish the hardware just did the whole position & tilt estimation, it would make things much easier. On my SP9 the current position estimate algorithm is not ideal, at times it seems to jump towards the edge of an antenna, particularly at some specific tilt angles. Seems less of a problem with other pens though, the Slim Pen 2 seems particularly susceptible to this problem.

StollD:
> * The lack of eraser functionality with my Chinese pen, but the only solution I see there is being able to let the eraser disable the button state. Thoughts on this?

This is more like a hot thought and not really thought through, but can you let one method guide the other one? For example, if the new method detects a button press, you ignore the button magnitude setting but do the same phase calculations to determine button / rubber? 

I don't know how reliable this would be though (because I don't really know the issue with the original method in the first place).

StollD:
From the SB2:

* With a rear mounted eraser, the button gets disabled in eraser mode.
* With an eraser button on the side, pressing the button disables the eraser.

src/core/generic/config.hpp
```diff
@@ -57,6 +57,9 @@ class Config {
 	usize dft_button_min_mag = 1000;
 	usize dft_freq_min_mag = 10000;
 	usize dft_tilt_min_mag = 10000;
+	usize dft_contact_min_mag = 10000;
```
iwanders, line 60: Should I prefix this with `dft_mpp2` as well? Probably right?

iwanders:
> I don't know how reliable this would be though (because I don't really know the issue with the original method in the first place).

Oh, mea culpa! I should've totally explained that in the description. So basically, the signal to noise ratio in the `DftButton` row 0 is kinda poor, for example in this [test](https://github.com/iwanders/analysis-ithc-iptsd/blob/aed699ffea03caec2fdba5d4c8a448928f4ce9f7/media/saturating_pressure/2024_02_22_slim_pen_pressure.png), the button isn't actually pressed, but the 0th row of the `DftButton` window clearly lights up. This is what causes the ghost button presses that slim pen 2 users have been reporting, I tried cranking the threshold up, as was suggested in one of the threads, while that resolved the ghost button presses it made the button detection only work when the tip was practically touching the screen for me. The 0th row is even more noisy at roughly 20 degrees towards the screen, and it is worse when the barrel button side is pointing towards the screen. I think it's still strong enough for the phase detection, I would've preferred a nice 'binary' signal, but maybe this is the best we have for now.

> This is more like a hot thought and not really thought through, but can you let one method guide the other one? For example, if the new method detects a button press, you ignore the button magnitude setting but do the same phase calculations to determine button / rubber?

Yep! I truly appreciate your suggestion, I think this works really well. I just pushed this in 2d49ff7, I changed `m_mppv2_button` into `m_mppv2_button_or_eraser`, then we can just use the phase check to discern between barrel button and eraser button. I got too stuck on the 'button detection' and didn't think of that we can also just use it as a 'button or eraser' detection.

This seems to work for all 4 pens I have here, all of them allow eraser detection as well as button detection without any changes to settings.

> but you want to press eraser and button at the same time?

Nope, you understood it right. I don't see much use in supporting both the button and the eraser. 


Current solution is nice; the new changes are all additive. If they cause problems they can also all be disabled by increasing the new magnitudes in the configuration. I don't expect problems of course, but it's nice to know that we have the knobs to tune if it does. I made one more comment on a variable name, I think that should be prefixed. I'll try to do some more testing this evening and over the next few days. Aiming to move this out of draft at the beginning of the weekend if I don't discover any issues during my testing.

We should also really consider making an application that just prints the relevant dft magnitudes, doesn't need to be pretty, but currently it is impossible to say - print the button magnitudes - without doing any code changes. Happy to file a PR for this if you're ok with something simple in this direction.

StollD:
> I tried cranking the threshold up, as was suggested in one of the threads, while that resolved the ghost button presses it made the button detection only work when the tip was practically touching the screen for me.

Thanks for the explanation! I figured that it must have had something to do with signal strength, but I didn't consider the distance from the screen.

> This seems to work for all 4 pens I have here, all of them allow eraser detection as well as button detection without any changes to settings.

Nice!

> We should also really consider making an application that just prints the relevant dft magnitudes, doesn't need to be pretty, but currently it is impossible to say - print the button magnitudes - without doing any code changes. Happy to file a PR for this if you're ok with something simple in this direction.

I have been thinking about how I would do a DFT calibration tool for a while, but I never got around to working on it because of other stuff. A simple debugging app that just prints the magnitudes would be a great start in that regard.

src/core/generic/config.hpp
StollD: Maybe I am blind ... but is that variable even used anywhere? You already have `dft_mpp_contact_min_mag`.

iwanders: No, you are not, I've removed it in 4895549b95654930869240708e677bfe4a0203fe probably a remnant of me renaming the variables :grimacing: , sorry about that!

iwanders:
Marked this as ready for review, it's good to go in imo: Did some more testing yesterday evening and Wednesday evening.  The code we arrived at seems to work well for all 4 pens.

StollD (approved):

quo:
Amazing work! I never even considered that the type 0xa windows would contain a button signal, I assumed it was a pen ID or something, since it just seemed to cycle through the same three values (and I don't have a MPP2 pen myself).

Code looks good to me. My only concern is that if the first type 0xa DFT window is missing for some reason, you will incorrectly read the button signal from the second window. So it would be better to just store the first 0xa DFT window, and delay the processing until you see the second one, so you know you're using the right window.

BTW, any chance you could share your raw data captures? When I have some more time I may try to figure out what the other 0xa bits mean.
(Ideally separate captures of hovering, contact, eraser hover, eraser contact, and button press for each pen.)

iwanders:
> BTW, any chance you could share your raw data captures? When I have some more time I may try to figure out what the other 0xa bits mean.

Certainly @quo , I've added a lot of it just now in a release associated to my analysis repository: https://github.com/iwanders/analysis-ithc-iptsd/releases/tag/v0.0.1

I hope you'll find that sufficient, it's the raw IRPMon recordings from Windows, my repo contains scripts to convert those to the iptsd binary files.

I think the `0xa` bits identify the pens with some unique id. I'm not sure how relevant that is. What currently still feels a bit wonky for the SP2 is the actual position estimates, in particular with the pen at an angle.

Another (probably more fun) thing to look at would be the way the pressure is encoded digitally, I looked at it briefly [here](https://github.com/iwanders/analysis-ithc-iptsd?tab=readme-ov-file#iptsdftwindowpressure), it does seem to contain a delta in each frame. I didn't figure it out though.

Lots of interesting things in this data, but ultimately I had to prioritize fixing the button glitching over exploring all the other information as time as limited. I'll have to step away from the Surface stuff to do some other things in my spare time, but if you do make some discoveries do ping me!

PR #186, opened 2025-02-16
Implement DFT partial update bundling

AndyGozas:
Because of the way DFT packets work, not all of the stylus data gets updated at the same time.

Previously, iptsd would just send out a new event every time a DFT packet arrived. Because of this several events containing the same position, pressure, etc. would be sent out.

This confused the smoothing algorithms in some applications, e.g. Krita, as described [here](https://github.com/linux-surface/linux-surface/issues/1588).

This PR implements bundling of these partial updates together into one single event.

### DIFF
--- a/etc/iptsd.conf
+++ b/etc/iptsd.conf
@@ -159,3 +159,4 @@
 # PositionExp = -0.7
 # ButtonMinMag = 1000
 # FreqMinMag = 10000
+# AllowSplitEvents = false
--- a/src/core/generic/application.hpp
+++ b/src/core/generic/application.hpp
@@ -233,8 +233,15 @@ private:
 		if (!m_info.is_touchscreen())
 			return;
 
-		m_dft.input(data);
-		this->process_stylus(m_dft.get_stylus());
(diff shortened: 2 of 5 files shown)


StollD (approved):

StollD:
Thank you!

As a side note, I wonder if an explicit sync after every data frame that iptsd reads from the hardware would work too (here: https://github.com/linux-surface/iptsd/blob/master/src/ipts/parser.hpp#L252), since one frame can contain multiple DFT windows.

AndyGozas:
> Thank you!
> 
> As a side note, I wonder if an explicit sync after every data frame that iptsd reads from the hardware would work too (here: https://github.com/linux-surface/iptsd/blob/master/src/ipts/parser.hpp#L252), since one frame can contain multiple DFT windows.

Not sure, but I doubt it. From what I've seen, it seems that all of the DFT data comes significantly staggered (and even it didn't, fixing it this way seems more complex as there would be more changes needed further down the line as well). I guess I'll know more whenever I get around to actually analyzing the DFT packets' contents.

PR #187, opened 2025-02-16
PKGBUILD: add /etc/iptsd.conf to backup

AndyGozas:
This stops pacman from deleting users' configs that have changes in them whenever the package is updated.


StollD (approved):

StollD:
Thank you!

PR #204, opened 2025-12-26
Fix typo in configuration loader breaking contact aspect logic

LegendaryFire:
There was a typo in the configuration loader causing the `AspectMin` configuration option to be set to the value of `AspectMax`. Fixed the typo and seems to be working as intended now.


qzed (approved):
Looks good to me, thanks!

PR #214, opened 2026-05-15
[Feature] Toggle touchscreen at different stylus levels

zeroxoneafour:
Palm rejection was bugging out on my surface, so I whipped up this patch to make it better.

You can now configure the touch screen to turn off -
* When it detects the stylus at all (this is the old default)
* Only when it detects that the stylus is being hovered or is being used
* Only when the stylus is being used

The new config setting, `Touchscreen.DisableOnStylusLevel`, allows you to pick between the three options. By default, for compatibility, `connected` is picked. This is also the case if the option is not there or is invalid. I personally prefer `active`, and `contact` guarantees stability when writing.

### DIFF
--- a/etc/iptsd.conf
+++ b/etc/iptsd.conf
@@ -20,7 +20,11 @@
 # DisableOnPalm = false
 
 ##
-## Ignore all touchscreen inputs if a stylus is in proximity.
+## Ignore all touchscreen inputs on a chosen stylus condition
+## default - Do not disable stylus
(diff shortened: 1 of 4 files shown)


zeroxoneafour:
I tested this on my Surface Pro 9 running Fedora 44 and it works. Idk why the CI runner for the Arch build failed.

hmtheboy154:
You might wanna squash all commits into 1

